PR #2, opened 2017-06-13
Storage location

smerritt:


timuralp:
Let's close this and reopen when we circle back to the storage location idea.

Issue #12, opened 2017-12-12
README container tags not consistent


clayg:
I think in [Trying it out](https://github.com/swiftstack/swift-s3-sync#trying-it-out) the container name (tag?) `swift-s3-sync` should maybe be `cloud-sync/test` to keep it consistent with the [integration test instructions](https://github.com/swiftstack/swift-s3-sync#integration-tests)?

Although now that I'm looking at the try it out section more closely, maybe the `docker run` command in try-it-out is already inconsistently with the build command?

PR #13, opened 2017-12-12
use swift-s3-sync container name consistently

clayg:
fixes issue #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ filesystem and uses a Swift all-in-one docker container as the base.
 The container must map the swift-s3-sync source tree when launching:
 
 ```
-docker run -P -d -v `pwd`:/swift-s3-sync cloud-sync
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,13 +93,13 @@ This middleware should be in the pipeline before the DLO/SLO middleware.
 ### Trying it out
 
 To try this out, you can build the docker container used for integration tests:
-`docker build -t swift-s3-sync test/container`. It uses S3Proxy, backed by the
+`docker build -t cloud-sync/test test/container`. It uses S3Proxy, backed by the
```
timuralp: I think we should maybe go the other way, as the repository is named "swift-s3-sync" and "cloud-sync" is not really a name that comes up in the README. Or I guess we could rename the repo, but if we do, I would rename it to cloud-connector or something, as it already does more than just sync.

PR #14, opened 2017-12-12
ignore the vendored repos

clayg:
Unless there's another more obvious solution I'm missing?


clayg:
@timuralp do these updates to `.gitignore` this work for you?  How are you handling this currently?

.gitignore
```diff
@@ -0,0 +1,3 @@
+*.egg-info
+/botocore
```
timuralp: I don't see these in the root repository, but I do see them in the `src/` directory. How did you setup the dependencies? I think the updates I made to the `requirements.txt` file should make `pip` do the right thing?

clayg:
@timuralp helped me get to the bottom of how this issue shows up in my workflow

PR #16, opened 2017-12-12
Feature/keystone auth

chianingwang:
Add support keystone auth v2 and v3 for cloud sync.
- s3_sync/sync_swift.py (m)
- test/unit/test_sync_swift.py (m)

### DIFF
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -55,26 +55,28 @@ class SyncSwift(BaseSync):
                              'retries': 3}
         # default: swift, options: keystone_v2 and keystone_v3
         _authtype = self.settings.get('auth_type', 'swift')
+        if _authtype == 'keystone_v2':
+            if 'tenant_name' not in self.settings:
+                raise ValueError(
+                    'Missing tenant_name for keystone v2')
+            connection_kwargs['tenant_name'] =\
+                self.settings['tenant_name']
+            connection_kwargs['auth_version'] = '2'
+        elif _authtype == 'keystone_v3':
+            _os_auth_fields = ['project_name',
+                               'project_domain_name',
+                               'user_domain_name']
+            missing_fields = [field for field in _os_auth_fields
+                              if field not in self.settings]
+            if missing_fields:
+                raise ValueError(
+                    'Missing %s for keystone v3' %
+                    ', '.join(missing_fields))
+            connection_kwargs['os_options'] = dict(
+                [(field, self.settings[field]) for field in _os_auth_fields])
+            connection_kwargs['auth_version'] = '3'
 
         def swift_client_factory():
-            if _authtype == 'keystone_v2':
-                try:
-                    _tenant_name = self.settings['tenant_name']
-                except Exception as e:
-                    raise Exception('missing_v2_arguments: ' + e.message)
-                connection_kwargs['tenant_name'] = _tenant_name
-                connection_kwargs['auth_version'] = '2'
-            elif _authtype == 'keystone_v3':
-                try:
-                    _os_auth_fields = ['project_name',
-                                       'project_domain_name',
-                                       'user_domain_name']
-                    _os_options = dict([(field, self.settings[field])
-                                        for field in _os_auth_fields])
-                except Exception as e:
-                    raise Exception('missing_v3_arguments: ' + e.message)
-                connection_kwargs['os_options'] = _os_options
-                connection_kwargs['auth_version'] = '3'
             return swiftclient.client.Connection(**connection_kwargs)
         return swift_client_factory
 
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -48,9 +48,34 @@ class SyncSwift(BaseSync):
         # auth URL
         endpoint = self.settings['aws_endpoint']
 
+        # **connection_args is common arguments; authurl, user, and key
+        connection_kwargs = {'authurl': endpoint,
+                             'user': username,
+                             'key': key,
+                             'retries': 3}
+        # default: swift, options: keystone_v2 and keystone_v3
+        _authtype = self.settings.get('auth_type', 'swift')
+
         def swift_client_factory():
-            return swiftclient.client.Connection(
-                authurl=endpoint, user=username, key=key, retries=3)
+            if _authtype == 'keystone_v2':
+                try:
+                    _tenant_name = self.settings['tenant_name']
+                except Exception as e:
+                    raise Exception('missing_v2_arguments: ' + e.message)
+                connection_kwargs['tenant_name'] = _tenant_name
+                connection_kwargs['auth_version'] = '2'
+            elif _authtype == 'keystone_v3':
+                try:
+                    _os_auth_fields = ['project_name',
+                                       'project_domain_name',
+                                       'user_domain_name']
+                    _os_options = dict([(field, self.settings[field])
+                                        for field in _os_auth_fields])
+                except Exception as e:
+                    raise Exception('missing_v3_arguments: ' + e.message)
+                connection_kwargs['os_options'] = _os_options
+                connection_kwargs['auth_version'] = '3'
+            return swiftclient.client.Connection(**connection_kwargs)
         return swift_client_factory
 
     def upload_object(self, name, policy, internal_client):
--- a/test/unit/test_sync_swift.py
+++ b/test/unit/test_sync_swift.py
@@ -581,3 +581,121 @@ class TestSyncSwift(unittest.TestCase):
         self.sync_swift.upload_object('foo', 'policy', mock_ic)
         self.assertEqual([mock.call.head_object('bucketcontainer', 'foo')],
                          swift_client.mock_calls)
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v2_auth_missing_tenant_name(self, mock_swift):
+        mock_swift.return_value = mock.Mock()
+        # in this case, the "bucket" is actually the prefix
+        aws_bucket = 'sync_'
+        sync_swift = SyncSwift(
+            {'aws_bucket': aws_bucket,
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v2',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_tenant_name = sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
+            missing_tenant_name()
+        self.assertTrue(
+            'missing_v2_arguments: tenant_name' in str(context.exception))
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v2_auth_current_withouterror(self, mock_swift):
+        mock_swift.return_value = mock.Mock()
+        aws_bucket = 'sync_'
+        sync_swift = SyncSwift(
+            {'aws_bucket': aws_bucket,
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v2',
+             'tenant_name': 'tenantname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        v2_withouterror = sync_swift._get_client_factory()
+        v2_withouterror()
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v3_auth_missing_project_name(self, mock_swift):
+        mock_swift.return_value = mock.Mock()
+        aws_bucket = 'sync_'
+        sync_swift = SyncSwift(
+            {'aws_bucket': aws_bucket,
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v3',
+             'user_domain_name': 'userdomainname',
+             'project_domain_name': 'projectdomainname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_project_name = sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
+            missing_project_name()
+        self.assertTrue(
+            'missing_v3_arguments: project_name' in str(context.exception))
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v3_auth_missing_project_domain_name(self, mock_swift):
+        aws_bucket = 'sync_'
+        sync_swift = SyncSwift(
+            {'aws_bucket': aws_bucket,
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v3',
+             'project_name': 'projectname',
+             'user_domain_name': 'userdomainname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_project_domain_name = sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
+            missing_project_domain_name()
+        self.assertTrue(
+            'missing_v3_arguments: project_domain_name'
+            in str(context.exception))
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v3_auth_missing_user_domain_name(self, mock_swift):
+        aws_bucket = 'sync_'
+        sync_swift = SyncSwift(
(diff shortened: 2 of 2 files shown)


timuralp:
@chianingwang could you rebase on master, now that we have travis hooked up, so we can run the unit/integration/flake8 tests?

chianingwang:
@timuralp I rebase and it looks like auto trigger the Travis test job but it looks like fail which can't get the container ID. I might miss something and might consult you tomorrow when I am in SF office.
```
$ container_id=$(mktemp)
before_install.3
0.27s
$ docker run -d -v `pwd`:/swift-s3-sync cloud-sync/test > "${container_id}"
6.64s
$ docker exec -it `cat ${container_id}` timeout 30 bash -c 'until echo > /dev/tcp/localhost/8080; do sleep 0.5; done'
bash: connect: Cannot assign requested address
bash: /dev/tcp/localhost/8080: Cannot assign requested address```

**Comment by chianingwang** [comment]
@timuralp I found the error and fix it. They are a most likely pem8 format issues.

One thing I don't know how to deal with which is as below.
Here is original code but it complains `'project_domain_name': self.settings['project_domain_name'],` line length 80 is longer than limitation 79.
```
            elif authtype == 'keystone_v3':
                _os_options = {
                  'user_domain_name': self.settings['user_domain_name'],
                  'project_domain_name': self.settings['project_domain_name'],
                  'project_name': self.settings['project_name']}
```

Thus I have to update the code like as below to deal with this issue.
```
            elif authtype == 'keystone_v3':
                userdomainname = self.settings['user_domain_name']
                projectdomainname = self.settings['project_domain_name']
                projectname = self.settings['project_name']
                _os_options = {
                    'user_domain_name': userdomainname,
                    'project_domain_name': projectdomainname,
                    'project_name': projectname}
```

I'm not sure whether there has a better way to handle this, if yes, pleases advice.

**Comment by timuralp** [comment]
@chianingwang I think that's fine. Alternatively, if you rewrite using the `dict` syntax, you could save some characters, as well. For example:
```
_os_options = dict(
    user_domain_name=self.settings['user_domain_name'],
    project_domain_name=self.settings['project_domain_name'],
    project_name=self.settings['project_name']
```

Of course, since the fields map in the config file map exactly to the code, you could even do this:
```
os_auth_fields = ['project_name', 'project_domain_name', 'user_domain_name']
_os_options = dict([(field, self.settings[field]) for field in os_auth_fields])
```

I don't have a strong preference around this, so you can pick whichever you prefer.

At the high level, can you please squash your commits and remove any commented out code? I hesitate to review this PR in its current state. Once that's done, I can start looking at the code.

**Comment by chianingwang** [comment]
@timuralp Thanks for your advice, I took the 2nd advice from you which I'm doing as below and looks good and pass the `Travis CI`
```
            elif authtype == 'keystone_v3':
                _os_auth_fields = ['project_name',
                                              'project_domain_name',
                                              'user_domain_name']
                _os_options = dict([(field, self.settings[field])
                                               for field in _os_auth_fields])
```
Moreover, I squash your commits and remove any commented out code.
Now the commit is only as below. I can squeeze them more or remove those comment totally. 
```
        # **connection_args is common arguments; authurl, user, and key
        common_connection_kwargs = {'authurl': endpoint,
                                    'user': username,
                                    'key': key,
                                    'retries': 3}

        # default: swift, options keystone_v2 and keystone_v3
        authtype = self.settings.get('auth_type', None)
        if authtype is None:
            authtype = 'swift'
```

**Comment by timuralp** [comment]
@chianingwang thanks for cleaning up some of the comment stuff! Can you squash the actual commits, please? There are currently 19 commits in this PR. Ideally, we would have one commit that implements the change required to support the keystone authentication option. Let me know if squashing commits with git is confusing to you -- happy to point you in the right direction!

File: s3_sync/sync_swift.py [hunk 1]
```diff
@@ -48,9 +48,40 @@ def _get_client_factory(self):
...
+        # default: swift, options keystone_v2 and keystone_v3
+        _authtype = self.settings.get('auth_type', None)
+        if _authtype is None:
+            _authtype = 'swift'
+
         def swift_client_factory():
-            return swiftclient.client.Connection(
-                authurl=endpoint, user=username, key=key, retries=3)
+            if _authtype == 'keystone_v2':
+                _tenant_name = self.settings['tenant_name']
```
timuralp: Consider changing this to a `.get()`. This way we won't hit a `KeyError`

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,40 @@ def _get_client_factory(self):
...
+                if _tenant_name:
+                    return swiftclient.client.Connection(
+                        auth_version='2',
+                        tenant_name=_tenant_name,
+                        **common_connection_kwargs)
+                else:
+                    raise NotImplementedError()
```
timuralp: Is this the right exception to raise? We do implement support for this auth type, but are missing arguments.

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,40 @@ def _get_client_factory(self):
...
+        # **connection_args is common arguments; authurl, user, and key
+        common_connection_kwargs = {'authurl': endpoint,
+                                    'user': username,
+                                    'key': key,
+                                    'retries': 3}
+
+        # default: swift, options keystone_v2 and keystone_v3
+        _authtype = self.settings.get('auth_type', None)
+        if _authtype is None:
+            _authtype = 'swift'
```
timuralp: Why not put `'swift'` as the argument to `.get()` on line 58?

test/unit/test_sync_swift.py
```diff
@@ -36,7 +36,8 @@ def setUp(self):
              'aws_secret': 'credential',
              'account': 'account',
              'container': 'container',
-             'aws_endpoint': 'http://swift.url/auth/v1.0'})
+             'aws_endpoint': 'http://swift.url/auth/v1.0',
+             'auth_type': 'swift'})
```
timuralp: Is this necessary? A missing argument should default to `swift` anyway?

test/unit/test_sync_swift.py
```diff
@@ -556,6 +557,7 @@ def test_per_account_bucket(self, mock_swift):
              'aws_secret': 'credential',
              'account': 'account',
              'container': 'container',
+             'auth_type': 'swift',
```
timuralp: Same question as above.

test/unit/test_sync_swift.py
```diff
@@ -581,3 +583,43 @@ def test_per_account_container_create(self, mock_swift):
         self.sync_swift.upload_object('foo', 'policy', mock_ic)
         self.assertEqual([mock.call.head_object('bucketcontainer', 'foo')],
                          swift_client.mock_calls)
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v2_auth(self, mock_swift):
+        mock_swift.return_value = mock.Mock()
+
+        # in this case, the "bucket" is actually the prefix
```
timuralp: I think this is an out of place artifact from a copy/paste.

test/unit/test_sync_swift.py
```diff
@@ -581,3 +583,43 @@ def test_per_account_container_create(self, mock_swift):
...
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v2',
+             'tenant_name': 'tenantname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+
+        self.assertEqual('sync_container', sync_swift.remote_container)
```
timuralp: You should assert that the `swiftclient.client.Connection` constructor was called with the expected arguments. Also, why do we want to assert the container name?

test/unit/test_sync_swift.py
```diff
@@ -581,3 +583,43 @@ def test_per_account_container_create(self, mock_swift):
...
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v3_auth(self, mock_swift):
+        mock_swift.return_value = mock.Mock()
+
+        # in this case, the "bucket" is actually the prefix
```
timuralp: Same comments as on the above test.

test/unit/test_sync_swift.py
```diff
@@ -581,3 +583,43 @@ def test_per_account_container_create(self, mock_swift):
...
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v3',
+             'user_domain_name': 'userdomainname',
+             'project_domain_name': 'projectdomainname',
+             'project_name': 'projectname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+
+        self.assertEqual('sync_container', sync_swift.remote_container)
```
timuralp: Same comments about the assertion.

chianingwang:
@timuralp, Thanks for helping me the code review and those are all great advice! I modified the code and commit it. Please help to review it and any advice will be welcome.

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,41 @@ def _get_client_factory(self):
...
+                    _os_auth_fields = ['project_name',
+                                       'project_domain_name',
+                                       'user_domain_name']
+                    _os_options = dict([(field, self.settings[field])
+                                        for field in _os_auth_fields])
+                    return swiftclient.client.Connection(
+                        auth_version='3',
+                        os_options=_os_options,
+                        **common_connection_kwargs)
+                except Exception as e:
```
timuralp: This is confusing to me. Why do we rescue the exception and then re-raise it? It seems that this could be removed entirely?

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,41 @@ def _get_client_factory(self):
...
-            return swiftclient.client.Connection(
-                authurl=endpoint, user=username, key=key, retries=3)
+            if _authtype == 'keystone_v2':
+                try:
+                    _tenant_name = self.settings['tenant_name']
+                    return swiftclient.client.Connection(
+                        auth_version='2',
+                        tenant_name=_tenant_name,
+                        **common_connection_kwargs)
+                except Exception as e:
```
timuralp: This is confusing to me. Why do we rescue the exception and then re-raise it? It seems that this could be removed entirely?

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,41 @@ def _get_client_factory(self):
...
+        # default: swift, options: keystone_v2 and keystone_v3
+        _authtype = self.settings.get('auth_type', 'swift')
+
         def swift_client_factory():
-            return swiftclient.client.Connection(
-                authurl=endpoint, user=username, key=key, retries=3)
+            if _authtype == 'keystone_v2':
+                try:
+                    _tenant_name = self.settings['tenant_name']
+                    return swiftclient.client.Connection(
```
timuralp: Ok, re-reading this, I think we can clean up the code a little more. In the `if` statements, we should only mutate the `common_connection_kwargs` dictionary (not actually calling `swiftclient.client.Connection`). And then have exactly one call to the method on line 85.

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,41 @@ def _get_client_factory(self):
...
+                                        for field in _os_auth_fields])
+                    return swiftclient.client.Connection(
+                        auth_version='3',
+                        os_options=_os_options,
+                        **common_connection_kwargs)
+                except Exception as e:
+                    raise e
+            else:
+                return swiftclient.client.Connection(
+                    authurl=endpoint, user=username, key=key, retries=3)
```
timuralp: This is exactly `common_connection_kwargs`. In fact, maybe naming it `connection_kwargs` (or `connection_args`) would be better.

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,75 @@ def test_per_account_container_create(self, mock_swift):
...
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v2',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_tenant_name = self.sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
+            missing_tenant_name()
+        self.assertTrue('tenant_name' in str(context.exception))
```
timuralp: Could we have a somewhat more specific message to check in the exception?

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,75 @@ def test_per_account_container_create(self, mock_swift):
...
+            {'aws_bucket': aws_bucket,
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v2',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_tenant_name = self.sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
```
timuralp: Is there a more specific exception than `Exception`?

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,75 @@ def test_per_account_container_create(self, mock_swift):
...
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v3',
+             'project_name': 'projectname',
+             'user_domain_name': 'userdomainname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_project_domain_name = self.sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
```
timuralp: We should have a more specific exception to check here (than `Exception`).

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,75 @@ def test_per_account_container_create(self, mock_swift):
...
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v3',
+             'user_domain_name': 'userdomainname',
+             'project_domain_name': 'projectdomainname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_project_name = self.sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
```
timuralp: We should have a more specific exception to check here (than `Exception`).

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,75 @@ def test_per_account_container_create(self, mock_swift):
         self.sync_swift.upload_object('foo', 'policy', mock_ic)
         self.assertEqual([mock.call.head_object('bucketcontainer', 'foo')],
                          swift_client.mock_calls)
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v2_auth_missing_argument(self, mock_swift):
```
timuralp: At the high level, can you write a test that asserts that the correct parameters are passed to the Connection() constructor? As is, we only have negative tests (tests that test for failure).

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,75 @@ def test_per_account_container_create(self, mock_swift):
...
+             'auth_type': 'keystone_v2',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'},
+            per_account=True)
+        missing_tenant_name = self.sync_swift._get_client_factory()
+        with self.assertRaises(Exception) as context:
+            missing_tenant_name()
+        self.assertTrue('tenant_name' in str(context.exception))
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_keystone_v3_auth_missing_arguments(self, mock_swift):
```
timuralp: This test has a lot of copy/pasted code. Can you improve on that? For example, the only thing that changes between each invocation is the missing parameter, right? We also need a test that verifies that when all parameters are present, we call the `Connection` class with the right arguments. Can you please write such a test?

chianingwang:
@timuralp  Great input and really appreciated the code review.
I modified the codes follow your suggestions but not sure I cover all of them.
Here is the summary what I changed. Please share your thought and any comment will be welcome.

##### s3_sync/sync_swift.py
 * rename `common_connection_kwargs` to `connection_kwargs`
 * handle `connection_kwargs` dict first and add required arguments base on `auth_type`
 * return connection at the end against `connection_kwargs`
 * more speficifc exception message via `raise Excpetion` 

##### test/unit/test_sync_swift.py
 * test_keystone_v2_auth_missing_tenant_name
 * test_keystone_v3_auth_missing_project_name
 * test_keystone_v3_auth_missing_project_domain_name
 * test_keystone_v3_auth_missing_user_domain_name
 * test_keystone_v2_auth_current_withouterror
 * test_keystone_v3_auth_current_withouterror
 
ps: I squash all the commits into 1 ! :+1:

chianingwang:
@timuralp I followed your advice and practice it, and run the unit test to test it. It's all good for me.

timuralp:
Closing this PR as we merged code based on it already and have keystone support. Thanks for spearheading this, @chianingwang!

PR #18, opened 2017-12-19
install s3_sync in container

clayg:


clayg:
@timuralp so for me I think this would be more convenient!  Do you wanna make sure it doesn't mess with your workflow?

test/container/launch.sh
```diff
@@ -33,17 +33,18 @@ if [ 0 -ne $? ]; then
...
 EOF
 fi
 
 set -e
 
-/usr/bin/sudo -u swift PYTHONPATH=/swift-s3-sync /swift/bin/startmain
+cd /swift-s3-sync; pip install -e .
 
-PYTHONPATH=/opt/ss/lib/python2.7/dist-packages:/swift-s3-sync \
-    python -m s3_sync --log-level debug \
```
clayg: @timuralp but *this* also removes the PYTHONPATH ???

PR #19, opened 2017-12-20
Remove PYTHONPATH in lieu of .pth.

timuralp:
Use a .pth file in /usr/lib/python2.7/dist-packages to avoid having to
rely on PYTHONPATH. This allows to still pick up code changes on
restarting daemons, but no longer requires using the PYTHONPATH
environment variable.


clayg:
is this better than #18 - or same?  or orthogonal?

timuralp:
@clayg I think it's better because it removes the `PYTHONPATH` stuff, but we can still restart the daemons to pick up new code.

timuralp:
I didn't understand what `pip install -e` does. We can merge that patch.

PR #20, opened 2017-12-21
Add get_object and head_object to the BaseSync.

timuralp:
We need to be able to issue GET requests to the remote containers not
just as part of shunting requests, but also when migrating objects
back into Swift. This patchs expands the existing API, such that the
migrator could LIST and GET all of the objects from the remote store
and stream them into Swift.

### DIFF
--- a/s3_sync/sync_s3.py
+++ b/s3_sync/sync_s3.py
@@ -28,6 +28,7 @@ import urllib
 from swift.common.internal_client import UnexpectedResponse
 from swift.common.utils import FileLikeIter
 from .base_sync import BaseSync
+from .base_sync import ProviderResponse
 from .utils import (
     convert_to_s3_headers, convert_to_swift_headers, FileWrapper,
     SLOFileWrapper, get_slo_etag, check_slo, SLO_ETAG_FIELD, SLO_HEADER,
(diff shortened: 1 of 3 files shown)


s3_sync/sync_s3.py
```diff
@@ -184,41 +185,62 @@ def shunt_object(self, req, swift_key):
...
+            'head_object', Bucket=self.aws_bucket, Key=key, **options)
+        response.body = ['']
+        return response
+
+    def get_object(self, key, **options):
+        response = self._call_boto(
+            'get_object', Bucket=self.aws_bucket, Key=key, **options)
+        if response.success:
+            old_body = response.body
+            response.body = iter(lambda: old_body.read(65536), b'')
```
tipabu: This makes me worry about our default value for `resp.get('Body', [''])` below -- if `Body` is missing, won't this fall over on an `AttributeError`?

timuralp: Yea, I need to think about this a little bit, as I had to add the default lookup because HEAD responses don't include a body. I think using `StringIO` here might be the right thing to do.

s3_sync/sync_swift.py
```diff
@@ -171,49 +172,64 @@ def shunt_object(self, req, name):
...
-                                     req.method)
-
+                    headers = resp
+                    body = ['']
                 status = 206 if 'content-range' in headers else 200
-                headers = [translate(header, value)
-                           for header, value in headers.items()]
-                return status, headers, body_iter
+                headers = dict([translate(header, value)
+                                for header, value in headers.items()])
```
tipabu: You're right -- of course it's already a dict! Yeah, never mind that comment.

s3_sync/sync_s3.py
```diff
@@ -184,41 +186,62 @@ def shunt_object(self, req, swift_key):
...
-                    body_iter = iter(lambda: resp['Body'].read(65536), b'')
-                else:
-                    resp = s3_client.head_object(**kwargs)
-                    body_iter = ['']
+                resp = getattr(s3_client, op)(**args)
+                return ProviderResponse(
+                    True, resp['ResponseMetadata']['HTTPStatusCode'],
+                    convert_to_swift_headers(
+                        resp['ResponseMetadata']['HTTPHeaders']),
+                    resp.get('Body', StringIO('')))
```
timuralp: @tipabu changed to a StringIO, which I think should work fine here?

tipabu: I was a little worried at first that we might be breaking the expectation for an iterable in `ProviderResponse` (in case we ever have some other function that wants to use `_call_boto` and eventually calls `to_wsgi` but *doesn't* do the `iter`-`lambda`-`read` trick), but I think we're probably good?

```
>>> [x for x in StringIO(u'asdf')]
['asdf']
```

Though -- is `botocore.response.StreamingBody` iterable? Should we just do a

```
if 'Body' in resp:
    body = iter(lambda: resp['Body'].read(65536), '')
else:
    body = ['']
return ProviderResponse(..., body)
```

here? We could keep the overwriting in head_object, too, just to be safe, but this way there's a more-definite contract for what `ProviderResponse.body` should look like.

timuralp: Yep, I'm on board. When we use this in the migrator, this may be useful as well.

timuralp:
@tipabu made the changes from the latest comment.

PR #22, opened 2017-12-21
Factor out the setup code from __main__.py.

timuralp:
As we plan to have additional daemons, we should factor out the shared
code from __main__.py to be reused elsewhere. Namely, we can reuse the
argument parsing, logger initialization, and loading the
configuration.

### DIFF
--- a/s3_sync/__main__.py
+++ b/s3_sync/__main__.py
@@ -14,94 +14,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-import argparse
-import json
 import logging
-import logging.handlers
 import os
-import sys
-import time
 import traceback
 
 from container_crawler import ContainerCrawler
-
-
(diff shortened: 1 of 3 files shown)


s3_sync/__main__.py
```diff
@@ -14,94 +14,17 @@
...
-                        help='path to the configuration file')
-    parser.add_argument('--once', action='store_true',
-                        help='run once')
-    parser.add_argument('--log-level', metavar='level', type=str,
-                        choices=['debug', 'info', 'warning', 'error'],
-                        help='logging level; defaults to info')
-    parser.add_argument('--console', action='store_true',
-                        help='log messages to console')
-    return parser.parse_args(args)
+from .daemon_utils import setup_context
```
clayg: cool!

clayg:
makes sense to me :shipit:

timuralp:
@clayg woot! merging!

PR #23, opened 2017-12-25
Add support for keystone auth with Swift.

timuralp:
Allows for using keystone v2 or v3 to connect to a Swift cluster. The
settings are exposed through the `auth_type` configuration parameter
for the container. The exposed options are:
- tenant_name (keystone v2)
- project_name, project_domain_name, user_domain_name (keystone v3)


timuralp:
@chianingwang here is a cleaned up version of your patch. Let me know what you think!

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,93 @@ def test_per_account_container_create(self, mock_swift):
...
+             'auth_type': 'keystone_v2',
+             'tenant_name': 'tenantname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'})
+        sync_swift._get_client_factory()()
+        mock_swift.assert_called_once_with(
+            authurl='http://swift.url/auth/v1.0', user='identity',
+            key='credential', auth_version='2',
+            tenant_name='tenantname', retries=3)
+
+    def test_keystone_v3_auth_missing_arguments(self):
```
timuralp: @chianingwang the missing arguments tests were mostly duplicated. I consolidated them into one test case with multiple parameters.

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,93 @@ def test_per_account_container_create(self, mock_swift):
...
+            'account': 'account',
+            'container': 'container',
+            'auth_type': 'keystone_v3',
+            'user_domain_name': 'userdomainname',
+            'project_name': 'projectname',
+            'project_domain_name': 'projectdomainname',
+            'aws_endpoint': 'http://swift.url/auth/v1.0'}
+        sync_swift = SyncSwift(settings)
+        sync_swift._get_client_factory()()
+        mock_swift.assert_called_once_with(
```
timuralp: @chianingwang this is what I meant by asserting that the correct arguments are passed to SwiftClient.

s3_sync/sync_swift.py
```diff
@@ -48,9 +48,36 @@ def _get_client_factory(self):
         # auth URL
         endpoint = self.settings['aws_endpoint']
 
+        # **connection_args is common arguments; authurl, user, and key
+        connection_kwargs = {'authurl': endpoint,
```
timuralp: I think we can pull all of this logic out of the factory method.

test/unit/test_sync_swift.py
```diff
@@ -581,3 +581,93 @@ def test_per_account_container_create(self, mock_swift):
...
+            {'aws_bucket': 'container',
+             'aws_identity': 'identity',
+             'aws_secret': 'credential',
+             'account': 'account',
+             'container': 'container',
+             'auth_type': 'keystone_v2',
+             'tenant_name': 'tenantname',
+             'aws_endpoint': 'http://swift.url/auth/v1.0'})
+        sync_swift._get_client_factory()()
+        mock_swift.assert_called_once_with(
```
timuralp: Here we're checking that the SwiftClient is called with the arguments we expect.

chianingwang:
@timuralp sorry for late response, it looks good. 
BTW, Thanks for your advice, I followed the way you shared and practiced in pull request #16, so either #16 or #23 you would like to merge are all fine for me.

timuralp:
@chianingwang check out the updated commits. I tried to add an integration test for v3 keystone auth. If all looks good, I think this patch would be reasonable to merge. We should add a v2 test -- I might circle back to that soon. I think it'd be ok to merge it without a v2 functional test, as it's a deprecated auth mechanism and @chianingwang did some functional testing. @chianingwang do you mind repeating the v2 functional test one more time with this branch?

PR #24, opened 2018-01-12
Example migration configuration.

timuralp:
For the migration work, add sample configuration for the migraiton
daemon. The assumption is that the swift-s3-sync main code and the
migrator will share the configuration (as well as the shunt).

Migrations are defined in their own array, however, some of the field
names must be preserved, so that they can be supplied to the
"provider" classes (i.e. SyncS3 and SyncSwift), as we will be relying
on them to issue GET for objects and listing the containers (and the
accounts, in case of Swift).


timuralp:
@clayg here is a sample configuration for the migrator. I realized that some of the fields we should keep the same, as there are assumptions about the configuration in the `SyncS3` and `SyncSwift` classes (which we want to use for GET/LIST requests).

PR #26, opened 2018-01-20
Return connections only after consuming the body.

timuralp:
When handling GET requests through the shunt, we should make sure the
body is consumed before returning the connection to the pool. This
means that we cannot use the context manager without turning the
get_object() function into a generator. The patch adds a wrapper
around the body that returns the connection to the pool after the body
is fully consumed (or the object is destroyed).

### DIFF
--- a/s3_sync/sync_s3.py
+++ b/s3_sync/sync_s3.py
@@ -31,8 +31,8 @@ from .base_sync import BaseSync
 from .base_sync import ProviderResponse
 from .utils import (
     convert_to_s3_headers, convert_to_swift_headers, FileWrapper,
-    SLOFileWrapper, get_slo_etag, check_slo, SLO_ETAG_FIELD, SLO_HEADER,
-    SWIFT_USER_META_PREFIX)
+    SLOFileWrapper, ClosingResourceIterable, get_slo_etag, check_slo,
+    SLO_ETAG_FIELD, SLO_HEADER, SWIFT_USER_META_PREFIX)
 
 
 class SyncS3(BaseSync):
@@ -219,14 +219,10 @@ class SyncS3(BaseSync):
             'get_object', Bucket=self.aws_bucket, Key=key, **options)
 
     def _call_boto(self, op, **args):
-        with self.client_pool.get_client() as boto_client:
-            s3_client = boto_client.client
+        def _perform_op(s3_client):
             try:
                 resp = getattr(s3_client, op)(**args)
-                if 'Body' in resp:
-                    body = iter(lambda: resp['Body'].read(65536), b'')
-                else:
-                    body = ['']
+                body = resp.get('Body', [''])
                 return ProviderResponse(
                     True, resp['ResponseMetadata']['HTTPStatusCode'],
                     convert_to_swift_headers(
@@ -237,10 +233,27 @@ class SyncS3(BaseSync):
                     False, e.response['ResponseMetadata']['HTTPStatusCode'],
                     convert_to_swift_headers(
                         e.response['ResponseMetadata']['HTTPHeaders']),
-                    e.response['Error']['Message'])
+                    iter(e.response['Error']['Message']))
             except Exception:
                 self.logger.exception('Error contacting remote s3 cluster')
-                return ProviderResponse(False, 502, {}, ['Bad Gateway'])
+                return ProviderResponse(False, 502, {}, iter('Bad Gateway'))
+
+        if op == 'get_object':
+            entry = self.client_pool.get_client()
+            resp = _perform_op(entry.client)
+            if resp.success:
+                resp.body = ClosingResourceIterable(
+                    entry,
+                    resp.body,
+                    length=int(resp.headers['Content-Length']))
+            else:
+                resp.body = ClosingResourceIterable(
+                    entry, resp.body, lambda: None)
+            return resp
+        else:
+            with self.client_pool.get_client() as boto_client:
+                s3_client = boto_client.client
+                return _perform_op(s3_client)
 
     def list_objects(self, marker, limit, prefix, delimiter):
         if limit > 1000:
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -23,7 +23,8 @@ import traceback
 
 from .base_sync import BaseSync
 from .base_sync import ProviderResponse
-from .utils import (FileWrapper, check_slo, SWIFT_USER_META_PREFIX)
+from .utils import (FileWrapper, ClosingResourceIterable, check_slo,
+                    SWIFT_USER_META_PREFIX)
 
 
 class SyncSwift(BaseSync):
@@ -208,10 +209,9 @@ class SyncSwift(BaseSync):
                 return ('Content-Length', value)
             return (header, value)
 
-        with self.client_pool.get_client() as client:
-            swift_client = client.client
+        def _perform_op(client):
             try:
-                resp = getattr(swift_client, op)(container, key, **args)
+                resp = getattr(client, op)(container, key, **args)
                 if isinstance(resp, tuple):
                     headers, body = resp
                 else:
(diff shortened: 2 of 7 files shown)


timuralp:
@tipabu do you know of a cleaner way to do this? I realized that I need to fix up before I make progress on the migrator code, as I need to make sure connections are not re-used prematurely.

s3_sync/utils.py
```diff
@@ -382,6 +382,42 @@ def read(self, size=-1):
...
+    def __iter__(self):
+        return self
+
+    def __del__(self):
+        """
+            There could be an exception raised, the resource is not put back in
+            the pool, and the content is not consumed. We handle this in the
+            destructor.
+        """
+        for _ in self.iterable:
```
tipabu: How big can this be? Would it be better to mark the resource as dirty in some way, close the socket without reading, and require that the pool create a fresh one?

s3_sync/utils.py
```diff
@@ -382,6 +382,42 @@ def read(self, size=-1):
...
+        iterable is consumed. The resource is assumed to implement the __exit__
+        method and can be used with a context manager.
+    """
+    def __init__(self, resource, iterable):
+        self.closed = False
+        self.iterable = iterable
+        self.resource = resource
+
+    def next(self):
+        try:
```
tipabu: I was kinda expecting a

    if self.closed:
        raise StopIteration

around here... Or maybe a `ValueError`? Something to say, "You tried to read from a thing that's already cleaned up. Don't do that."

s3_sync/utils.py
```diff
@@ -382,6 +382,42 @@ def read(self, size=-1):
...
+            return next(self.iterable)
+        except StopIteration:
```
tipabu: Should this catch more than just `StopIteration`?

If calling `next(self.iterable)` raises some *other* error (likely a socket error, I guess?) should we still call `self.resource.close()`?

s3_sync/utils.py
timuralp: Good point -- will add this!

s3_sync/utils.py
timuralp: Good point. We should. I need to check what `swiftclient` and `boto` do on socket errors, but presumably they can continue using the session and we should still return it to the pool.

s3_sync/utils.py
timuralp: This is tough. I need to check if there is a way to do this with `boto` or `swiftclient` (which probably will involve looking into what plumbing to requests/urllib3 they expose). Your point makes sense -- I'd be for it. The connection re-creation/marking the resource as dirty is also not supported with the current code -- would need to expand that a little bit.

timuralp:
@tipabu please re-review this if you get a chance. I added a bunch of tests for the resource tracking, to make sure connections are returned to the pool when we expect them to be.

s3_sync/sync_s3.py
```diff
@@ -219,14 +219,10 @@ def get_object(self, key, **options):
...
-        with self.client_pool.get_client() as boto_client:
-            s3_client = boto_client.client
+        def _perform_op(s3_client):
             try:
                 resp = getattr(s3_client, op)(**args)
-                if 'Body' in resp:
-                    body = iter(lambda: resp['Body'].read(65536), b'')
-                else:
-                    body = ['']
+                body = resp.get('Body', [''])
```
tipabu: Still weirds me out a little that we can pass `ProviderResponse` any one of

- an iterable with a `close()` method,
- an iterable _without_ a `close()` method, or
- a readable

I guess... as long as it works? But it feels like we're making `to_wsgi()` a bit of a lie, if only briefly...

And wait, shouldn't `ProviderResponse` get a `close()` method? Oh, I see -- we're updating the body and passing `resp.body` to `ClosingResourceIterable`... hmm...

s3_sync/sync_swift.py
```diff
@@ -226,10 +226,25 @@ def translate(header, value):
...
                 self.logger.exception('Error contacting remote swift cluster')
-                return ProviderResponse(False, 502, {}, ['Bad Gateway'])
+                return ProviderResponse(False, 502, {}, iter('Bad Gateway'))
+
+        if op == 'get_object':
+            client = self.client_pool.get_client()
+            resp = _perform_op(client.client)
+            if resp.success:
+                resp.body = ClosingResourceIterable(
+                    client, resp.body, resp.body.resp.close)
```
tipabu: > `resp.body.resp.close`

Not confusing at all! 😛 Did you ever get around to writing that swiftclient patch? I wanna +2 something! I wanna +2 something!

s3_sync/utils.py
```diff
@@ -382,6 +382,80 @@ def read(self, size=-1):
...
+        self.resource = resource
+        self.length = length
+        self.amt_read = 0
+
+        if close_callable:
+            self.close_data = close_callable
+        elif hasattr(self.data_src, 'close'):
+            self.close_data = self.data_src.close
+        else:
+            raise ValueError('No closeable to close the data source defined')
```
tipabu: I wonder if it'd be better to just pass -- then we could get rid of the `lambda: None`s above.

s3_sync/utils.py
```diff
@@ -382,6 +382,80 @@ def read(self, size=-1):
...
+
+        try:
+            iter(self.data_src)
```
tipabu: Don't we want

    self.data_src = iter(self.data_src)

? Otherwise, when we pass `['']` on `HEAD`s, `data_src` will still be a list and we'll hit something like

    >>> data_src=['']
    >>> next(data_src)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    TypeError: list object is not an iterator

no?

s3_sync/utils.py
```diff
@@ -382,6 +382,80 @@ def read(self, size=-1):
...
+            else:
+                if self.amt_read == self.length:
+                    raise StopIteration
+                ret = self.data_src.read(self.read_chunk)
+                self.amt_read += len(ret)
+                if not ret:
+                    raise StopIteration
+                return ret
+        except Exception as e:
+            if type(e) != StopIteration:
```
tipabu: Not

    if isinstance(e, StopIteration):

? Eh, who am I kidding -- no one really subclasses *that*, right?

test/unit/test_sync_swift.py
```diff
@@ -26,17 +26,42 @@
 from utils import FakeStream
 
 
+class FakeBody(object):
```
tipabu: I'm a little surprised this didn't get a `close()`, maybe something like https://github.com/openstack/swift/blob/2.16.0/test/unit/common/middleware/helpers.py#L30-L41 ?

s3_sync/sync_s3.py
timuralp: This part is a little confusing -- agreed. The `ProviderResponse` is only a tiny wrapper that allows me to avoid passing tuples around. It doesn't really do anything clever with the body and the `to_wsgi()` call only returns the 3 tuples so that we can use it in the `return` statement.

s3_sync/sync_swift.py
timuralp: I'll do it this weekend! Agreed!

s3_sync/utils.py
timuralp: Maybe. I could check if the `close_data` is defined or not. I can rework that bit.

s3_sync/utils.py
timuralp: The `HEAD` code path does not end up using the `ClosingResourceIterable`. I only use it for the `get_object` calls. However, in that code path, we can have errors passed in, but I return them as `iter()`, which should be ok? It'd be nice if there was a better way to test WSGI handling of errors here -- I'll try to write a functional test for it.

s3_sync/utils.py
timuralp: Done

test/unit/test_sync_swift.py
timuralp: That's a good idea! I probably should check for this as well.

PR #27, opened 2018-01-24
Add swift-s3-sync-verify command

tipabu:
... so you can verify your remote credentials more easily. Usage:

```
usage: swift-s3-verify [-h] --protocol {s3,swift} --endpoint ENDPOINT --username
                       USERNAME --password PASSWORD [--bucket BUCKET]
```

This will use the provided creds to go create an object in the remote cloud, update metadata, HEAD the object, DELETE the object, and list the bucket. If any of that fails, exit 1 with a (hopefully helpful) message.

Note that `BUCKET` *can* be `/*` (but only when `--protocol` is `swift`), in which case we'll create (and then remove) a container named `.cloudsync_test_container-👍` before doing the above.

<s>I should really, *really* add an integration test for this -- we've already got an s3proxy instance spun up...</s>


setup.py
```diff
@@ -17,6 +17,7 @@
       entry_points={
           'console_scripts': [
               'swift-s3-sync = s3_sync.__main__:main'
+              'swift-s3-sync-verify = s3_sync.verify:main'
```
timuralp: Could we name it `swift-s3-verify`? The migrator will be `swift-s3-migrator`, so we're sort of trying to have the `swift-s3` as the higher level prefix (to be changed at some point to a better name).

s3_sync/verify.py
```diff
@@ -0,0 +1,122 @@
...
+            else:
+                swift_key = 'cloud_sync_test_object'
+            validate_bucket(provider, swift_key,
+                            args.protocol == 'swift' and args.bucket == '/*')
+    except botocore.exceptions.ClientError as e:
+        if e.response['Error']['Code'] == 'SignatureDoesNotMatch':
+            exit('Invalid credentials. Please check the Access Key '
+                 'ID and Secret Access Key.')
+        else:
+            exit('Unexpected error validating credentials: %r' % (
```
timuralp: One common error could be the bucket not existing, but that's not as big of a deal if we don't specifically call it out.

timuralp:
@tipabu awesome! This looks pretty good. An integration test would actually be pretty straightforward to add, but we can do it later. One other follow on change would be to add a way to only validate that we can `LIST` buckets/containers. Currently, we use that to verify that credentials are plausibly valid.

tipabu:
> One other follow on change would be to add a way to only validate that we can `LIST` buckets/containers. Currently, we use that to verify that credentials are plausibly valid.

We've got that, actually -- it's baked into `main()` around line 90. Just leave off `--bucket`.

timuralp:
Cool -- I obviously missed it.

setup.py
tipabu: Done.

timuralp:
Looks great! Thanks @tipabu!

PR #29, opened 2018-02-02
Clear endpoint when verifying against AWS

tipabu:
Otherwise, Boto starts complaining about a missing region?

### DIFF
--- a/s3_sync/verify.py
+++ b/s3_sync/verify.py
@@ -46,18 +46,42 @@ def validate_bucket(provider, swift_key, create_bucket):
     if create_bucket:
         # This should only be necessary on Swift; reach down to the client
         with provider.client_pool.get_client() as client:
-            client.client.put_container(provider.aws_bucket)
+            result = client.client.put_container(provider.aws_bucket)
+        if result is not None:
+            return result
+
     internal_client = FakeInternalClient('cloud sync test', {})
-    provider.upload_object(swift_key, 0, internal_client)
-    provider.update_metadata(swift_key, {'X-Object-Meta-Cloud-Sync': 'fabcab',
-                                         'content-type': 'text/plain'})
-    provider.head_object(swift_key)
-    provider.delete_object(swift_key)
-    provider.list_objects(marker='', limit=1, prefix='', delimiter='')
+    result = provider.upload_object(swift_key, 0, internal_client)
+    if result is not None:
+        return result
+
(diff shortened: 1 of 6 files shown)


s3_sync/verify.py
```diff
@@ -79,6 +80,8 @@ def main(args=None):
     }
     if args.bucket == '/*':
         conf['aws_bucket'] = u'.cloudsync_test_container-\U0001f44d'
+    if urlparse.urlparse(args.endpoint).hostname.endswith('.amazonaws.com'):
```
timuralp: This makes sense. I started skimming the `SyncS3` class and noticed that we do the following:
```
 64         if not self.endpoint or self.endpoint.endswith('amazonaws.com'):
 65             # We always use v4 signer with Amazon, as it will support all
 66             # regions.
 67             boto_config = boto3.session.Config(signature_version='s3v4',
 68                                                s3={'aws_chunked': True})
[..]
 78             s3_client = boto_session.client('s3',
 79                                             endpoint_url=self.endpoint,
 80                                             config=boto_config)
```

And I wonder if we should change the unconditional usage of the `endpoint_url` or ignore it for AWS. The right way to force usage of a specific region would be to pass the `region_name` parameter and rely on the boto3 region dictionary to figure it out. I don't think we should change that now, but there is a longer-term missing bit in supporting region-aware configurations. We're currently relying on boto to do a HEAD on the bucket, but it'd be nice to allow clients to skip that by supplying an explicit region name (I need to validate that it actually works in the sane way that I imagine).

timuralp:
@tipabu sorry for the delay! It'd be great to add a unit test for this, as well, if you get to it.

s3_sync/verify.py
```diff
@@ -46,18 +46,42 @@ def validate_bucket(provider, swift_key, create_bucket):
...
-    provider.update_metadata(swift_key, {'X-Object-Meta-Cloud-Sync': 'fabcab',
-                                         'content-type': 'text/plain'})
-    provider.head_object(swift_key)
-    provider.delete_object(swift_key)
-    provider.list_objects(marker='', limit=1, prefix='', delimiter='')
+    result = provider.upload_object(swift_key, 0, internal_client)
+    if result is not None:
+        return result
+
+    result = provider.update_metadata(swift_key, {
```
timuralp: This is my failing, but it'd be good if `update_metadata` would also return a `ProviderResponse` object :(

PR #31, opened 2018-02-07
Add poll_interval to swift-s3-migrator

clayg:
By default we'll cycle at most once every five seconds, you can
configure the poll_interval in the migration_settings section of the
json config.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -414,6 +414,45 @@ class Migrator(object):
                 self.object_queue.task_done()
 
 
+def process_migrations(migrations, migration_status, internal_pool, logger,
+                       items_chunk, node_id, nodes):
+    for index, migration in enumerate(migrations):
+        if migration['aws_bucket'] == '/*' or index % nodes == node_id:
+            try:
+                if migration.get('remote_account'):
+                    src_account = migration.get('remote_account')
+                else:
+                    src_account = migration['aws_identity']
+                logger.debug('Processing %s' % (
+                    ':'.join([migration.get('aws_endpoint', ''),
+                              src_account, migration['aws_bucket']])))
+                migrator = Migrator(migration, migration_status,
+                                    items_chunk,
+                                    internal_pool, logger,
+                                    node_id, nodes)
+                migrator.next_pass()
+            except Exception as e:
+                logger.error('Migration error: %r\n%s' % (
+                    e, traceback.format_exc(e)))
+
+
+def run(migrations, migration_status, internal_pool, logger, items_chunk,
+        node_id, nodes, poll_interval, once):
+    while True:
+        cycle_start = time.time()
+        process_migrations(migrations, migration_status, internal_pool, logger,
+                           items_chunk, node_id, nodes)
+        elapsed = time.time() - cycle_start
+        naptime = max(0, poll_interval - elapsed)
+        msg = 'Finished cycle in %0.2fs' % elapsed
+        if once:
+            logger.debug(msg)
+            return
+        msg += ', sleeping for %0.2fs.' % naptime
+        logger.debug(msg)
+        time.sleep(naptime)
+
+
 def main():
     args, conf = setup_context(
         description='Daemon to migrate objects into Swift')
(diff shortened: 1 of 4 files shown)


s3_sync/migrator.py
```diff
@@ -414,6 +414,45 @@ def _upload_worker(self):
...
+                logger.error('Migration error: %r\n%s' % (
+                    e, traceback.format_exc(e)))
+
+
+def run(migrations, migration_status, internal_pool, logger, items_chunk,
+        node_id, nodes, poll_interval, once):
+    while True:
+        cycle_start = time.time()
+        process_migrations(migrations, migration_status, internal_pool, logger,
+                           items_chunk, node_id, nodes)
```
clayg: I think passing around all this state could improve further - Swift adopted the pattern of making each Daemon a class, doing all the config parsing in __init__ and then calling run, which has a number of benefits beyond just familiarity.

test/container/swift-s3-sync.conf
```diff
@@ -75,10 +75,19 @@
             "restore_object": true
         }
     ],
+    "migrations": [],
```
clayg: this was enough for me to functionally test *this* change 😉

test/unit/test_migrator.py
```diff
@@ -90,3 +95,63 @@ def test_listing_comparison(self):
...
+    @contextmanager
+    def patch(self, name):
+        with mock.patch('s3_sync.migrator.' + name) as mocked:
+            yield mocked
+
+    def pop_log_lines(self):
+        lines = self.stream.getvalue()
+        self.stream.seek(0)
+        self.stream.truncate()
+        return lines
```
clayg: Swift has built up a fairly sophisticated debug logger fake over the years, I'm weary of the prospect of rebuilding all that adhoc - maybe there's a py2 backport of assertLogs [1] ???

1. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs

s3_sync/migrator.py
```diff
@@ -414,6 +414,45 @@ def _upload_worker(self):
...
 
+def process_migrations(migrations, migration_status, internal_pool, logger,
+                       items_chunk, node_id, nodes):
+    for index, migration in enumerate(migrations):
+        if migration['aws_bucket'] == '/*' or index % nodes == node_id:
+            try:
+                if migration.get('remote_account'):
+                    src_account = migration.get('remote_account')
+                else:
+                    src_account = migration['aws_identity']
```
tipabu: nit: Maybe better as

    src_account = migration.get('remote_account', migration['aws_identity'])

?

This is just for logging, right? We don't do anything like this in Migrator? Otherwise, how would a reseller-admin user `alice` migrate the account at `/v1/alice` instead of `/v1/AUTH_alice`?

s3_sync/migrator.py
```diff
@@ -414,6 +414,45 @@ def _upload_worker(self):
...
+
+def run(migrations, migration_status, internal_pool, logger, items_chunk,
+        node_id, nodes, poll_interval, once):
+    while True:
+        cycle_start = time.time()
+        process_migrations(migrations, migration_status, internal_pool, logger,
+                           items_chunk, node_id, nodes)
+        elapsed = time.time() - cycle_start
+        naptime = max(0, poll_interval - elapsed)
+        msg = 'Finished cycle in %0.2fs' % elapsed
```
tipabu: Wouldn't mind seeing this get dropped to a recon file as well, so I don't have to go parsing logs to track progress.

Do we collect any stats about how much has moved? That'd also be nice...

test/container/swift-s3-sync.conf
```diff
@@ -75,10 +75,19 @@
...
+    "migrator_settings": {
+        "items_chunk": 1000,
+        "log_file": "/var/log/swift-s3-migrator.log",
+        "poll_interval": 1,
```
tipabu: Out of curiosity, why not just `interval`, to match swift?

test/container/swift-s3-sync.conf
```diff
@@ -75,10 +75,19 @@
...
+        "status_file": "/var/lib/swift-s3-sync/migrator.status",
```
tipabu: Is this used yet?

timuralp: Yes! You can totally read one of these on your migrator-enabled node!

s3_sync/migrator.py
timuralp: We plan to, but not today :\

s3_sync/migrator.py
timuralp: Yep, only used for logging. The logic that handles this storage URL override is buried in the `SyncSwift` class:
```
55         if self.settings.get('remote_account'):
56             scheme, rest = endpoint.split(':', 1)
57             host = urllib.splithost(rest)[0]
58             path = '/v1/%s' % urllib.quote(self.settings['remote_account'])
59             os_options = {
60                 'object_storage_url': '%s:%s%s' % (scheme, host, path)}
```

timuralp:
I'm going to merge this and defer renaming `poll_interval` to `interval` -- I'm happy to do it, but we should also then rename the `container-crawler` poll interval option to match (just for consistency sake).

PR #34, opened 2018-02-13
Shunt bimodal accounts

tipabu:
Be able to archive ProxyFS data to a remote cloud. Presumably sync will also work, though its utility is questionable.

Since we walk container listings to find data that needs moving, we store the actual log segment data in the remote cloud (as opposed to a snapshot of the current file hierarchy). This requires that the shunt middleware be placed *after* the pfs middleware. However, we *also* have special handling around static large objects that requires the shunt be placed *before* slo -- resolve this by introducing a new config option (`handle_proxyfs`) and updating the proxy-server config to something like

```
[pipeline:main]
pipeline = ... cloud_shunt ... slo ... pfs pfs_cloud_shunt ... proxy-server

...

[filter:cloud_shunt]
use = egg:swift-s3-sync#cloud-shunt
conf_file = ...
handle_proxyfs = false

[filter:pfs_cloud_shunt]
use = egg:swift-s3-sync#cloud-shunt
conf_file = ...
handle_proxyfs = true
```

The no-auth pipeline used by proxyfsd will also need to be updated; FWIW, I'm running:

```
[pipeline:main]
pipeline = catch_errors gatekeeper healthcheck proxy-logging cache copy dlo meta pfs_cloud_shunt versioned_writes proxy-logging proxy-server

...

[filter:pfs_cloud_shunt]
use = egg:swift-s3-sync#cloud-shunt
conf_file = ...
handle_proxyfs = true
```

(I'm *pretty* sure we want to be after `meta`?)

**Note**: *only* log segment data should be archived -- checkpoint data we want to stay local.

Requires https://github.com/swiftstack/ProxyFS/pull/74 to prevent the sync daemon from thinking that a log segment containing an SLO manifest needs to have its segments synced. (They'd all 404.)

Requires https://github.com/swiftstack/container-crawler/pull/6 to allow us to skip `.__checkpoint__` containers.

Probably also requires https://github.com/swiftstack/ProxyFS/pull/72 ?

### DIFF
--- a/s3_sync/utils.py
+++ b/s3_sync/utils.py
@@ -525,3 +525,25 @@ def check_slo(swift_meta):
     if SLO_HEADER not in swift_meta:
         return False
     return swift_meta[SLO_HEADER].lower() == 'true'
+
+
+def response_is_complete(status_code, headers):
+    if status_code == 200:
+        return True
+    if status_code != 206:
+        return False
+    # Let's see if that Partial Content really is "partial"
+    cr = ''
+    for header, value in headers:
+        if header.lower() == 'content-range':
+            cr = value
+            break
+    if not cr.startswith('bytes 0-'):
+        return False
+    end, length = cr[8:].partition('/')[::2]
+    try:
+        end = int(end)
+        length = int(length)
+    except ValueError:
+        return False
+    return end + 1 == length
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ pip install swift-s3-sync-<version>.tar.gz
 
 You also will need to install the `container-crawler` library from Git:
 ```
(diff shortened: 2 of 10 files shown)


timuralp:
@tipabu we'll need to update the container crawler dependency before this will pass the tests. I'll start there.

s3_sync/utils.py
```diff
@@ -525,3 +525,25 @@ def check_slo(swift_meta):
     if SLO_HEADER not in swift_meta:
         return False
     return swift_meta[SLO_HEADER].lower() == 'true'
+
+
+def response_is_complete(status_code, headers):
```
timuralp: Adding a test for this might be a good idea.

PR #36, opened 2018-02-15
Add an integration test for the migrator.

timuralp:
Adds two tests for the migrator, which verify moving simple objects from
S3 and Swift.

### DIFF
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -1,0 +1,108 @@
+"""
+Copyright 2018 SwiftStack
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import StringIO
+from . import (TestCloudSyncBase, clear_swift_container, clear_s3_bucket,
+               wait_for_condition)
+
+
+class TestMigrator(TestCloudSyncBase):
+    @classmethod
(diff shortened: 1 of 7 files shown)


test/integration/test_migrator.py
```diff
@@ -0,0 +1,102 @@
+"""
+Copyright 2017 SwiftStack
```
tipabu: It's 2018 now :P

test/integration/test_migrator.py
```diff
@@ -0,0 +1,102 @@
...
+               wait_for_condition)
+
+
+class TestMigrator(TestCloudSyncBase):
+    @classmethod
+    def _find_migration(klass, matcher):
+        for migration in klass.test_conf['migrations']:
+            if matcher(migration):
+                return migration
+        raise RuntimeError('No matching migration')
```
tipabu: Maybe better as a `raise SkipTest`?

test/integration/test_migrator.py
```diff
@@ -0,0 +1,102 @@
...
+
+        wait_for_condition(5, _check_objects_copied)
+
+        for name, expected_body, user_meta in test_objects:
+            hdrs, body = self.local_swift(
+                'get_object', migration['container'], name)
+            self.assertEqual(body, body)
+            for k, v in user_meta.items():
+                self.assertIn('x-object-meta-' + k, hdrs)
+                self.assertEqual(v, hdrs['x-object-meta-' + k])
```
tipabu: Should we have assertions about what's in the remote? Do we leave objects, or delete them as we move them?

test/integration/test_migrator.py
timuralp: Oh, doh! Do we have to go through and twiddle the years on all of these?! :(

test/integration/test_migrator.py
timuralp: I like this because it told me that my configuration file was busted. Would prefer to keep that behavior.

test/integration/test_migrator.py
timuralp: We let them be -- I can add assertions about that.

PR #39, opened 2018-02-20
Migrate container ACLs in Swift->Swift configurations

timuralp:
Aside from just metadata, we should also propagate container ACLs during the migration.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -33,6 +33,7 @@ from .provider_factory import create_provider
 from .utils import (convert_to_local_headers, convert_to_swift_headers,
                     SWIFT_TIME_FMT)
 from swift.common.internal_client import UnexpectedResponse
+from swift.common.swob import Request
 from swift.common.utils import FileLikeIter, Timestamp
 
 
@@ -282,13 +283,24 @@ class Migrator(object):
                 raise MigrationError('Failed to HEAD bucket/container %s' %
                                      container)
             headers = {}
+            acl_hdrs = ['x-container-read', 'x-container-write']
             for hdr in resp.headers:
-                if hdr.startswith('x-container-meta-'):
+                if hdr.startswith('x-container-meta-') or hdr in acl_hdrs:
                     headers[hdr] = resp.headers[hdr]
         else:
             headers = {}
-        internal_client.create_container(
-            self.config['account'], container, headers)
+
+        req = Request.blank(
+            internal_client.make_path(self.config['account'], container),
+            environ={'REQUEST_METHOD': 'PUT',
+                     'swift_owner': True},
+            headers=headers)
+
+        resp = req.get_response(internal_client.app)
+        if resp.status_int // 100 != 2:
+            raise UnexpectedResponse('Failed to create container %s: %d' % (
+                container, resp.status_int), resp)
+
         start = time.time()
         while time.time() - start < timeout:
             if not internal_client.container_exists(
--- a/test/integration/__init__.py
+++ b/test/integration/__init__.py
@@ -26,9 +26,19 @@ import unittest
 
 
 def clear_swift_container(client, container):
-    _, list_results = client.get_container(container)
+    try:
+        _, list_results = client.get_container(container)
+    except swiftclient.exceptions.ClientException as e:
+        if e.http_status == 404:
+            return
+        raise
     for obj in list_results:
-        client.delete_object(container, obj['name'])
+        try:
+            client.delete_object(container, obj['name'])
+        except swiftclient.exceptions.ClientException as e:
+            if e.http_status == 404:
+                continue
+            raise
 
 
 def clear_s3_bucket(client, bucket):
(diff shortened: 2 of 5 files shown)


test/integration/test_migrator.py
```diff
@@ -106,3 +108,37 @@ def _check_objects_copied():
...
+                    return False
+                raise
+
+        hdrs, listing = wait_for_condition(5, _check_container_created)
+        self.assertIn('x-container-meta-test', hdrs)
+        self.assertEqual('test metadata', hdrs['x-container-meta-test'])
+
+        scheme, rest = self.SWIFT_CREDS['authurl'].split(':', 1)
+        swift_host, _ = urllib.splithost(rest)
+        remote_local_swift = swiftclient.client.Connection(
```
clayg: I lost track of this test at this variable name, `remote_local_` seems paradoxical.

test/unit/test_migrator.py
```diff
@@ -537,13 +537,17 @@ def test_migrate_objects_reset(self, create_provider_mock):
 
     @mock.patch('s3_sync.migrator.create_provider')
     def test_missing_container(self, create_provider_mock):
-        tests = [{'protocol': 'swift'}, {}]
+        tests = [
+            ({'protocol': 'swift'}, {'x-container-meta-foo': 'foo'}),
+            ({}, {}),
+            ({'protocol': 'swift'}, {'x-container-read': '.r*',
+                                     'x-container-write': 'AUTH_bob'})
+        ]
```
clayg: this seems like good infrastructure

s3_sync/migrator.py
```diff
@@ -282,13 +283,24 @@ def _create_container(self, container, internal_client, timeout=1):
...
                     headers[hdr] = resp.headers[hdr]
         else:
             headers = {}
-        internal_client.create_container(
-            self.config['account'], container, headers)
+
+        req = Request.blank(
+            internal_client.make_path(self.config['account'], container),
+            environ={'REQUEST_METHOD': 'PUT',
+                     'swift_owner': True},
```
clayg: acl's get filtered otherwise huh, interesting...

I'm concerned that integration tests still pass with this diff:

```
diff --git a/s3_sync/migrator.py b/s3_sync/migrator.py
index 65692f7..07e9774 100644
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -33,7 +33,6 @@ from .provider_factory import create_provider
 from .utils import (convert_to_local_headers, convert_to_swift_headers,
                     SWIFT_TIME_FMT)
 from swift.common.internal_client import UnexpectedResponse
-from swift.common.swob import Request
 from swift.common.utils import FileLikeIter, Timestamp
 
 
@@ -290,16 +289,8 @@ class Migrator(object):
         else:
             headers = {}
 
-        req = Request.blank(
-            internal_client.make_path(self.config['account'], container),
-            environ={'REQUEST_METHOD': 'PUT',
-                     'swift_owner': True},
-            headers=headers)
-
-        resp = req.get_response(internal_client.app)
-        if resp.status_int // 100 != 2:
-            raise UnexpectedResponse('Failed to create container %s: %d' % (
-                container, resp.status_int), resp)
+        internal_client.create_container(self.config['account'], container,
+                                         headers)
 
         start = time.time()
         while time.time() - start < timeout:
```

test/integration/__init__.py
```diff
@@ -26,7 +26,12 @@
...
 def clear_swift_container(client, container):
-    _, list_results = client.get_container(container)
+    try:
+        _, list_results = client.get_container(container)
+    except swiftclient.exceptions.ClientException as e:
+        if e.http_status == 404:
+            return
+        raise
     for obj in list_results:
         client.delete_object(container, obj['name'])
```
clayg: I think you can get bit by a 404 here as well?  If any individual item in the listing is out of date - still be nice to try the request on the rest of items in the list

test/integration/__init__.py
```diff
@@ -195,8 +204,12 @@ def _get_s3_sync_conf(self):
 
     @staticmethod
     def _remove_swift_container(client, container):
-        clear_swift_container(client, container)
-        client.delete_container(container)
+        try:
+            clear_swift_container(client, container)
```
clayg: I think `clear_swift_container` should be handling all it's own 404's and moved out this of this try

test/integration/test_migrator.py
timuralp: I could not think of a good name for a storage URL override connection. Will keep thinking.

s3_sync/migrator.py
timuralp: Thanks for the pointer -- will investigate!

test/integration/__init__.py
timuralp: True -- I'll clean that up as part of this as well.

test/integration/__init__.py
timuralp: Done

PR #40, opened 2018-02-21
Remove unconfigured migrations status on startup

clayg:
Any stale status entries for migrations which are no longer configured
on the node will be pruned during startup of s3-sync-migrator.


test/unit/test_migrator.py
```diff
@@ -281,6 +282,165 @@ def test_status_save_raises(self, mock_open):
...
+                "status": {
+                    "finished": 1519177178.41313,
+                    "last_finished": 1519177173.780246,
+                    "last_moved_count": 1,
+                    "last_scanned_count": 1,
+                    "marker": "blah",
+                    "moved_count": 1,
+                    "scanned_count": 1
+                }
+            }])
```
clayg: I find this test a lot easier to read and reason about - it gives me a much higher confidence that the code works.

test/unit/test_migrator.py
```diff
@@ -281,6 +282,165 @@ def test_status_save_raises(self, mock_open):
...
+            "status": {
+                "finished": 1519177178.41313,
+                "last_finished": 1519177173.780246,
+                "last_moved_count": 1,
+                "last_scanned_count": 1,
+                "marker": "blah",
+                "moved_count": 1,
+                "scanned_count": 1
+            }
+        }])
```
clayg: I don't like reading this test, and I had a lot of trouble writing it.  I had to crib a lot of ideas about how to avoid the filesystem from surrounding test infrastructure.  It doesn't seem like a great trade off of developer time, but perhaps I'm just more resolved to being less clever than my peers are.

PR #41, opened 2018-02-21
Do not duplicate names in listings.

timuralp:
If an object appears in both remote store and local swift, the shunt
should only display one entry, with a list for content_location, which
includes both places.

### DIFF
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -314,7 +314,8 @@ class S3SyncShunt(object):
                                 '<?xml version="1.0" encoding="UTF-8"?>', 1)
 
         # Default to plain format
-        return '\n'.join([entry['name'] for entry in list_results])
+        return u'\n'.join(entry['name'] if 'name' in entry else entry['subdir']
+                          for entry in list_results).encode('utf-8')
 
 
 def filter_factory(global_conf, **local_conf):
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -181,24 +181,41 @@ class S3SyncShunt(object):
                 cloud_name = None
 
             if cloud_name is not None:
+                added = False
+
                 if internal_name is None or \
                         (internal_name is not None and
-                         cloud_name <= internal_name):
+                         cloud_name < internal_name):
+                    resp[cloud_index]['content_location'] = [
+                        resp[cloud_index]['content_location']]
                     spliced_response.append(resp[cloud_index])
                     cloud_index += 1
-                    if len(resp) == cloud_index:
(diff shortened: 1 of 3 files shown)


s3_sync/shunt.py
```diff
@@ -181,24 +181,41 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+                added = False
+
                 if internal_name is None or \
                         (internal_name is not None and
-                         cloud_name <= internal_name):
+                         cloud_name < internal_name):
+                    resp[cloud_index]['content_location'] = [
+                        resp[cloud_index]['content_location']]
                     spliced_response.append(resp[cloud_index])
                     cloud_index += 1
```
clayg: so the "bug" as it were mostly had to do with in the < case we do *not* increment internal_index, but in the == case we *do*

s3_sync/shunt.py
```diff
@@ -141,66 +166,45 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
-                    cloud_name = resp[cloud_index]['subdir']
+            while remote_item and remote_key < local_key:
+                spliced_response.append(remote_item)
+                remote_item, remote_key = next(remote_iter)
+
+            if remote_key == local_key:
+                # duplicate!
+                remote_item['content_location'].append('swift')
+                spliced_response.append(remote_item)
+                remote_item, remote_key = next(remote_iter)
```
clayg: noice

s3_sync/shunt.py
```diff
@@ -297,7 +301,8 @@ def _format_listing_response(list_results, list_format, container):
                                 '<?xml version="1.0" encoding="UTF-8"?>', 1)
 
         # Default to plain format
-        return '\n'.join([entry['name'] for entry in list_results])
+        return u'\n'.join(entry['name'] if 'name' in entry else entry['subdir']
```
clayg: what `subdir`!?  we're pros now!

test/unit/test_shunt.py
```diff
@@ -636,4 +667,49 @@ def test_list_container_shunt_swift(self):
             mock.call('', 10000, '', ''),
             mock.call(u'unicod\xe9'.encode('utf-8'), 10000, '', '')])
         names = body_iter.split('\n')
-        self.assertEqual(['abc', u'unicod\xe9'], names)
+        self.assertEqual(['abc', u'unicod\xe9'.encode('utf-8')], names)
+
+    def test_list_container_shunt_with_duplicates(self):
```
clayg: that's what I'm looking for

PR #42, opened 2018-02-21
Do not store credentials in status files.

timuralp:
We should never write out cloud credentials in the status files.


test/unit/test_migrator.py
```diff
@@ -46,37 +46,46 @@ def test_migration_comparison(self):
...
-             False)]
+             False),
+            ({'account': 'swift',
+              'aws_bucket': 'bucket',
+              'aws_identity': 'aws access key',
+              'aws_secret': 'old secret'},
+             {'account': 'swift',
+              'aws_bucket': 'bucket',
+              'aws_identity': 'aws access key',
+              'aws_secret': 'new secret'},
```
clayg: noice

s3_sync/migrator.py
```diff
@@ -52,6 +52,8 @@ def equal_migration(left, right):
     for k in left.keys():
         if k == 'status':
             continue
+        if k == 'aws_secret':
+            continue
```
clayg: @FoxLisk once imparted on me the following wisdom:

> in computer science there are only three numbers: 0, 1 and N.

```
if k in IGNORED_KEYS: 
```

s3_sync/migrator.py
```diff
@@ -154,10 +156,13 @@ def save_migration(self, migration, marker, moved_count, scanned_count,
             if equal_migration(entry, migration):
                 if 'status' not in entry:
                     entry['status'] = {}
+                if 'aws_secret' in entry:
+                    del entry['aws_secret']
```
clayg: oh, man - that `IGNORED_KEYS` would be *super* handy!

test/unit/test_migrator.py
```diff
@@ -234,7 +243,11 @@ def test_status_save(self, mock_open):
             # gets the 1st argument in the call argument list
             written = ''.join([call[1][0] for call in
                                mock_file.write.mock_calls])
-            self.assertEqual(json.loads(written), [migration])
+            written_conf = json.loads(written)
+            for entry in written_conf:
+                self.assertTrue('aws_secret' not in entry)
+            del migration['aws_secret']
+            self.assertEqual(written_conf, [migration])
```
clayg: this is some belts and braces stuff right here...

s3_sync/migrator.py
```diff
@@ -154,10 +156,13 @@ def save_migration(self, migration, marker, moved_count, scanned_count,
...
                 status = entry['status']
                 break
         else:
             entry = dict(migration)
+            del entry['aws_secret']
```
clayg: This would be more robust and causes slightly less tests to fail as:

```
entry.pop('aws_secret', None)
```

test/unit/test_migrator.py
```diff
@@ -1022,7 +1041,7 @@ def test_update_status_finished_no_moves_resets_last(self):
         }, status)
 
     def test_update_status_clean_finish_does_not_reset_last(self):
-        # and we'll stay this way, indefinately...
+        # and we'll stay this way, indefinitely...
```
clayg: mixing concerns 😜 and also thanks 🙏

PR #43, opened 2018-02-23
Fix a bug in checking for equality of migrations.

clayg:
meh, I should fix the commit message


test/unit/test_migrator.py
```diff
@@ -43,53 +43,70 @@ def create_list_timestamp(epoch_ts):
...
+              'aws_identity': 'aws access key',
+              'status': {'moved': 100,
+                         'scanned': 200}},
+             True),
+            ({'account': 'swift1',
               'aws_bucket': 'bucket',
               'aws_identity': 'aws access key',
               'aws_secret': 'old secret'},
-             {'account': 'swift',
+             {'account': 'swift1',
```
clayg: i probably should have named these `swift2`

s3_sync/migrator.py
```diff
@@ -120,13 +120,16 @@ def __init__(self, status_location):
...
 
     def load_status_list(self):
+        self.status_list = []
         try:
             with open(self.status_location) as fh:
-                self.status_list = json.load(fh)
+                try:
+                    self.status_list = json.load(fh)
+                except ValueError as e:
+                    if str(e) != 'No JSON object could be decoded' == str(e):
```
clayg: there's no explicit test for this change

timuralp: Did you mean to not include the last `str(e)`? I'm confused by this otherwise.

PR #44, opened 2018-02-23
Add context in status to identify all bucket migrations

clayg:


s3_sync/migrator.py
```diff
@@ -224,6 +224,7 @@ def next_pass(self):
             self._next_pass()
             return
 
+        self.config['all_buckets'] = True
         self.config['container'] = '.'
```
timuralp: We can probably drop this then.

test/unit/test_migrator.py
```diff
@@ -407,12 +411,13 @@ def test_status_prune(self):
             "remote_account": "",
         }])
         with open(self.status_file_path) as rf:
-            self.assertEqual(json.load(rf), [{
+            self.assertEqual(json.load(rf), json.loads(json.dumps([{
```
timuralp: Was the comparison not stable otherwise?

PR #46, opened 2018-03-09
Do not double release a connection semaphore.

timuralp:
If we double release the connection semaphore, the semaphore count goes
to 2, resulting in the connection being available to be checked out
twice. This can lead to data corruption, as multiple co-routines may be
using the same connection.

The patch adds some tests and a check for the semaphore balance before
calling release to try to detect such bugs.

### DIFF
--- a/s3_sync/base_sync.py
+++ b/s3_sync/base_sync.py
@@ -54,6 +54,10 @@ class BaseSync(object):
             return self.semaphore.acquire(blocking=False)
 
         def close(self):
+            if self.semaphore.balance > BaseSync.HTTP_CONN_POOL_SIZE - 1:
+                logging.getLogger('s3-sync').error(
+                    'Detected double release of the semaphore')
+                raise RuntimeError('Detected double release of the semaphore!')
             self.semaphore.release()
             self.pool.release()
 
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -385,6 +385,7 @@ class Migrator(object):
             args['resp_chunk_size'] = 65536
         resp = self.provider.get_object(key, **args)
         if resp.status != 200:
+            resp.body.close()
             raise MigrationError('Failed to GET %s/%s: %s' % (
                 container, key, resp.body))
         put_headers = convert_to_local_headers(
--- a/s3_sync/utils.py
+++ b/s3_sync/utils.py
@@ -447,6 +447,7 @@ class ClosingResourceIterable(object):
         if not self.exhausted:
             self.close_data()
         if not self.closed:
+            self.closed = True
             self.resource.close()
 
     def __next__(self):
--- a/test/unit/test_utils.py
+++ b/test/unit/test_utils.py
@@ -18,6 +18,8 @@ limitations under the License.
 
 import mock
 from s3_sync import utils
+from s3_sync import base_sync
+import StringIO
 import unittest
 from utils import FakeStream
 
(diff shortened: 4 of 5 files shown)


test/unit/test_utils.py
```diff
@@ -178,3 +180,57 @@ def get_object(account, container, key, headers={}):
...
+
+    def test_resource_close_destructor(self):
+        pool = mock.Mock()
+        resource = base_sync.BaseSync.HttpClientPoolEntry(None, pool)
+        self.assertTrue(resource.acquire())
+        self.assertEqual(0, resource.semaphore.balance)
+        data_src = StringIO.StringIO('test data')
+        closing_iter = utils.ClosingResourceIterable(
+            resource, data_src)
+        del closing_iter
```
FoxLisk: huh. i did not expect `del` to force the destructor to run immediately. the docs indicate that it does https://docs.python.org/2/reference/simple_stmts.html#del 

> del x doesn’t directly call x.__del__() — the former decrements the reference count for x by one, and the latter is only called when x’s reference count reaches zero.

however, other resources[1] indicate that CPython actually uses a generational GC and I struggle to find anyone saying that when a reference count is dropped to 0, that object is immediately collected 

anyway, i'm assuming the test passes and thus the destructor is getting called in the way you expect and this is really academic. just surprising, at least to me.

[1] https://www.quora.com/How-does-garbage-collection-in-Python-work-What-are-the-pros-and-cons

s3_sync/base_sync.py
```diff
@@ -54,6 +54,8 @@ def acquire(self):
             return self.semaphore.acquire(blocking=False)
 
         def close(self):
+            if self.semaphore.balance > BaseSync.HTTP_CONN_POOL_SIZE - 1:
+                raise RuntimeError('Detected double release of the semaphore!')
```
FoxLisk: If this gets called during `__del__`, you will probably lose this exception. from https://docs.python.org/2/reference/datamodel.html?highlight=destructor#object.__del__ --

> Due to the precarious circumstances under which __del__() methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead.

timuralp: That is unfortunate, but I don't think we can do any better here (sadly).

FoxLisk: Maybe log it as well, so it doesnt get eaten by stderr -- at least then you'd have a record of it, if you end up looking through logs? not a lot of value there but arguably better than nothing

timuralp: Done!

s3_sync/migrator.py
```diff
@@ -385,6 +385,7 @@ def _migrate_object(self, container, key):
             args['resp_chunk_size'] = 65536
         resp = self.provider.get_object(key, **args)
         if resp.status != 200:
+            resp.body.close()
```
tipabu: Do we want something similar down around L427? Or should HEADs be safe, since spec requires that there be no body?

s3_sync/utils.py
```diff
@@ -461,7 +462,8 @@ def __del__(self):
             the pool, and the content is not consumed. We handle this in the
             destructor.
         """
-        self.close()
+        if not self.closed:
```
tipabu: Is the guard really necessary? I thought part of the point of L450 was that you could call `close()` repeatedly without worrying...

test/unit/test_utils.py
```diff
@@ -178,3 +180,57 @@ def get_object(account, container, key, headers={}):
...
+        self.assertTrue(resource.acquire())
+        self.assertEqual(0, resource.semaphore.balance)
+        data_src = StringIO.StringIO('test data')
+        closing_iter = utils.ClosingResourceIterable(
+            resource, data_src)
+        closing_iter.close()
+        self.assertEqual(1, resource.semaphore.balance)
+        self.assertTrue(closing_iter.closed)
+        del closing_iter
+        self.assertEqual(1, resource.semaphore.balance)
```
tipabu: Worth adding at least one test with double explicit `close()`s? Or an explicit `close()` after reading all data?

s3_sync/migrator.py
timuralp: Yep, only GET is special. For all other requests, we don't use the `ClosingResourceIterable` and buffer the body in memory.

s3_sync/utils.py
timuralp: Oh, you're right! Done.

test/unit/test_utils.py
timuralp: Done

PR #47, opened 2018-03-10
Close connections after each Migrator pass.

timuralp:
After each pass of the Migrator, we should close connections to the
remote cluster. Otherwise, we rely on HTTP timeouts, which can take
minutes, while creating more connections on each pass. This comes up
most acutely during tests, where the migrator can exhause the Swift
all-in-one connection pools on the proxy server.

### DIFF
--- a/s3_sync/base_sync.py
+++ b/s3_sync/base_sync.py
@@ -173,6 +173,12 @@ class BaseSync(object):
     def list_buckets(self):
         raise NotImplementedError()
 
+    def close(self):
+        for client in self.client_pool.client_pool:
+            client.acquire()
+            self._close_conn(client.client)
+            client.close()
+
     def _get_client_factory(self):
         raise NotImplementedError()
 
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -216,6 +216,7 @@ class Migrator(object):
         self.logger = logger
         self.node_id = node_id
         self.nodes = nodes
+        self.provider = None
 
     def next_pass(self):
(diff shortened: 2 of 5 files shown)


s3_sync/migrator.py
```diff
@@ -501,6 +508,7 @@ def process_migrations(migrations, migration_status, internal_pool, logger,
                                     internal_pool, logger,
                                     node_id, nodes)
                 migrator.next_pass()
+                migrator.close()
```
tipabu: Now I kinda want to add `__enter__`/`__exit__` methods to `Migrator` -- or replace the whole thing with a function 😕

s3_sync/base_sync.py
```diff
@@ -169,6 +169,11 @@ def list_objects(self, marker, limit, prefix, delimiter=None):
     def list_buckets(self):
         raise NotImplementedError()
 
+    def close(self):
+        for client in self.client_pool.client_pool:
+            with client as conn:
+                self._close_conn(conn)
```
tipabu: Where is this getting defined? Or are we just relying on subclasses implementing it on their own?

timuralp: The latter. This is different depending on the SDK, because neither boto3 nor swiftclient have an obvious way to close the requests session. It'd be pretty cool if `close` was a thing that did that -- at least that will soon be the case for python-swiftclient!

s3_sync/migrator.py
timuralp: Sure -- I'm going to defer that bit of refactoring to another time, though.

PR #48, opened 2018-03-10
Use the ResponseMetadata for status codes from S3.

timuralp:
The Error dictionary from S3 does not contain the HTTP status code. We
already use the ResponseMetadata from boto3 in all places and should do
so in list_objects, as well.

### DIFF
--- a/s3_sync/sync_s3.py
+++ b/s3_sync/sync_s3.py
@@ -331,7 +331,8 @@ class SyncS3(BaseSync):
                     keys + prefixes,
                     key=lambda x: x['name'] if 'name' in x else x['subdir']))
         except botocore.exceptions.ClientError as e:
-            return (e.response['Error']['Code'], e.message)
+            return (e.response['ResponseMetadata']['HTTPStatusCode'],
+                    e.message)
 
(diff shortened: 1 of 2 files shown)


s3_sync/sync_s3.py
```diff
@@ -326,7 +326,8 @@ def list_objects(self, marker, limit, prefix, delimiter=None):
                     keys + prefixes,
                     key=lambda x: x['name'] if 'name' in x else x['subdir']))
         except botocore.exceptions.ClientError as e:
-            return (e.response['Error']['Code'], e.message)
+            return (e.response['ResponseMetadata']['HTTPStatusCode'],
+                    e.message)
```
tipabu: Is `e.message` like `e.response['Error']['Message']` that we use up in `_call_boto`? Or will it be something else entirely?

timuralp: This is how botocore defines it:
```
msg = self.MSG_TEMPLATE.format(
            error_code=error_response['Error'].get('Code', 'Unknown'),
            error_message=error_response['Error'].get('Message', 'Unknown'),
            operation_name=operation_name)
```

So it is `['Error']['Message']` with the operation name and the error code (which is not the HTTP status code).

PR #49, opened 2018-03-10
Migrator: The metadata ETag check is backwards.

timuralp:
Fixes the ETag check, along with a unit test.

### DIFF
--- a/test/unit/test_migrator.py
+++ b/test/unit/test_migrator.py
@@ -538,6 +538,27 @@ class TestMigratorUtils(unittest.TestCase):
         status.load_status_list()
         self.assertEqual(status_list, status.status_list)
 
+    def test_cmp_meta(self):
+        test_cases = [
+            ({'last-modified': create_timestamp(1.5e9),
+              'etag': 'foo'},
+             {'last-modified': create_timestamp(1.5e9),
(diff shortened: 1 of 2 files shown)


test/unit/test_migrator.py
```diff
@@ -538,6 +538,27 @@ def test_load_status(self):
         status.load_status_list()
         self.assertEqual(status_list, status.status_list)
 
+    def test_cmp_meta(self):
+        test_cases = [
+            ({'last-modified': create_timestamp(1.5e9),
+              'etag': 'foo'},
+             {'last-modified': create_timestamp(1.5e9),
+              'etag': 'bar'},
+             s3_sync.migrator.EQUAL),
```
tipabu: Those don't *look* equal...

timuralp: Yea, that was somewhat on purpose -- the current assumption is that if the Last-Modified dates are equal, then we consider the objects equal and move on. I could add an additional ETag check, but there is really not much we can do when there is an ETag mismatch, but the Last-Modified dates are equal, as our update will be rejected since we set X-Timestamp. We could use the offset thing to make it work correctly, but I'd rather not tackle that yet.

PR #50, opened 2018-03-12
DLO support.

timuralp:
Support for DLO and a small pythonic improvement in the migrator.


timuralp:
I'm going to merge this in, as we'd like to pull it into the next version.

PR #56, opened 2018-03-21
Initial cleanup/refactoring for cloud-connector

dbishop:
Made method interfaces consistent between BaseSync abastract base class
and its children.

Pulled out `filter_hop_by_hop_headers` to utility file (cloud-connector
will want to use that).

Added plumbing in SyncS3.__init__() to take an AWS session token in the
settings (used when deploying containers to amazon ECS, when the
environment will have URL that responds with expiring, session-based
credentials).  In the cloud-connect use-case, the session is only used
once, so there's no provision for noticing credential expiration and
fetching new credentials from the URL.

Added `curl` to the integration test/development testing container,
since it's handy for manual testing.

### DIFF
--- a/s3_sync/sync_s3.py
+++ b/s3_sync/sync_s3.py
@@ -58,9 +58,14 @@ class SyncS3(BaseSync):
         aws_secret = self.settings['aws_secret']
         self.encryption = self.settings.get('encryption', True)
 
-        boto_session = boto3.session.Session(
-            aws_access_key_id=aws_identity,
-            aws_secret_access_key=aws_secret)
+        session_kwargs = {
+            'aws_access_key_id': aws_identity,
+            'aws_secret_access_key': aws_secret,
+        }
+        if self.settings.get('aws_session_token', None):
+            session_kwargs['aws_session_token'] = \
+                self.settings['aws_session_token']
+        boto_session = boto3.session.Session(**session_kwargs)
         if not self.endpoint or self.endpoint.endswith('amazonaws.com'):
             # We always use v4 signer with Amazon, as it will support all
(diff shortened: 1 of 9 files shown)


dbishop:
I even did a UNIT TEST?!

s3_sync/sync_s3.py
```diff
@@ -169,7 +174,7 @@ def _delete_not_found(self, s3_key):
                 else:
                     raise
 
-    def delete_object(self, swift_key, internal_client=None):
+    def delete_object(self, swift_key):
```
timuralp: Doesn't this break here: https://github.com/swiftstack/swift-s3-sync/blob/master/s3_sync/sync_container.py#L111? To be fair, if tests pass, we should have a test for such API changes.

s3_sync/sync_swift.py
```diff
@@ -113,42 +114,42 @@ def upload_object(self, name, policy, internal_client):
...
             swift_client.put_object(self.remote_container,
-                                    name,
+                                    swift_key,
                                     wrapper_stream,
                                     etag=wrapper_stream.get_headers()['etag'],
                                     headers=headers,
                                     content_length=len(wrapper_stream))
 
-    def delete_object(self, name, internal_client=None):
+    def delete_object(self, swift_key):
```
timuralp: Same comment as in sync_s3.

s3_sync/sync_s3.py
dbishop: Looked like no implementation of `delete_object` used that arg.

dbishop:
@timuralp the bad call-site is fixed

PR #57, opened 2018-03-22
We should preserve the ETag for Swift uploads.

timuralp:
When uploading objects into Swift, we used to replace the ETag header
with Content-MD5. As Swift ignores the Content-MD5, this is actually
wrong and we should rely on the ETag header.


timuralp:
John spotted this while reading the description of the migrator. Thanks for the catch, John!

mdegerne:
oops - was trying to push rebased branch. Opening new PR with recreated branch.

mdegerne:
If you found your way here, you should probably go to: https://github.com/swiftstack/swift-s3-sync/pull/64

PR #58, opened 2018-03-22
Honor Container Specification for Migration

mdegerne:
This will allow migration of a single container to a new location.

Notes:
 - SLO and DLO segment containers keep names across migration - no change for now

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -18,6 +18,7 @@ import eventlet
 import eventlet.pools
 eventlet.patcher.monkey_patch(all=True)
 
+from collections import namedtuple
 import datetime
 import errno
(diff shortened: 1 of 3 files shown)


s3_sync/migrator.py
```diff
@@ -463,15 +476,15 @@ def _find_missing_objects(
                         marker = remote['name']
         return marker, scanned, copied
 
-    def _migrate_object(self, container, key):
-        args = {'bucket': container}
+    def _migrate_object(self, aws_bucket, container, key):
```
timuralp: We can probably change the function signature to expect the named tuple? I wouldn't worry about that for this patch though. Happy to merge as is.

PR #59, opened 2018-03-23
First shot at a minimal cloud-connector daemon

dbishop:
The cloud-connector is designed to run in a container in a public cloud
whose "local" object storage uses the S3 API.  Its configuration is
stored in that "local" object storage and the daemon's container
requires enough configuration via environment variables to be able to
find its main config file, and a copy of the cloud-sync JSON config
file.

The idea is to provide a Swift or S3 API in the public cloud which has
"seamless" access to data in the on-prem Swift cluster.  The primary
use-case is bursting workloads to the public cloud which normally run in
private cloud, using the on-prem Swift cluster.

Further work will expand cloud-connector's capabilities and unlock more
use cases.  It will grow into its name.

This first version only supports a rudimentary GET and HEAD path for
objects.

It also needs tests.

But the infrastructure to be able to write integration tests against it
(and interact with it in the container via a bash shell) is there.


s3_sync/cloud_connector/app.py
```diff
@@ -0,0 +1,242 @@
...
+        super(CloudConnectorController, self).__init__(app)
+
+        self.version = version
+        self.account_name = account_name  # UTF8-encoded string
+        self.container_name = container_name  # UTF8-encoded string
+        self.object_name = object_name  # UTF8-encoded string
+        self.sync_profile, per_account = \
+            maybe_munge_profile_for_all_containers(sync_profile)
+        self.provider = create_provider(self.sync_profile, max_conns=5,
+                                        per_account=per_account)
```
clayg: nice, so no changes to the provider interface?

s3_sync/cloud_connector/util.py
```diff
@@ -0,0 +1,177 @@
...
+        'aws_bucket': env_options['CONF_BUCKET'],
+    }
+    # To default to real Amazon S3, the aws_endpoint kwarg needs to not be
+    # present.
+    if env_options['CONF_ENDPOINT']:
+        provider_settings['aws_endpoint'] = env_options['CONF_ENDPOINT']
+    if env_options.get('AWS_SECURITY_TOKEN_STRING', None):
+        provider_settings['aws_session_token'] = \
+            env_options['AWS_SECURITY_TOKEN_STRING']
+    provider = SyncS3(provider_settings)
```
clayg: IIUC, we're targeting AWS pretty specifically here, so this is fine... but I noticed above we used generic create_provider...

dbishop: Well, S3 API; that could be Amazon S3, some Google thing, or, I guess, a Swift cluster running `swift3`

s3_sync/cloud_connector/app.py
dbishop: i cut the spike's PUT path code; it was crappy anyway; one or more new methods will be coming, when PUT comes back.

PR #60, opened 2018-03-23
Handle Swift versioning in migrations.

timuralp:
When migrating containers, we should consider Swift versioning and
handle it correctly. There are two types: history and versions. One
behaves like a historical view of the container/objects and the other is
more akin to a stack (like S3 versioning), where objects are popped off
the stack as versions are deleted.


mdegerne:
Rebased with utf8 code fixes and made tests work.

dbishop:
lgtm, but I didn't set up versioned containers of each type, add some objs with multi versions, migrate, then make sure the versions work right in the new cluster... that'd be what I'd need to _really_ feel good.

PR #61, opened 2018-03-23
Allow a custom_prefix for s3 to be used

mdegerne:
Instead of using the MD5 hash function, allow a custom prefix
to be defined. THIS CAN CAUSE DATA LOSS if not used very carefully.

Migrate multiple containers to same S3 bucket with custom prefix
defined will cause over-write of matching object names.

[#155407966]


mdegerne:
Closing - This is not exactly what is called for in the ticket.

PR #63, opened 2018-03-27
Allow a custom_prefix for s3 to be used

mdegerne:
Instead of using the MD5 hash function, allow a custom prefix
to be defined. THIS CAN CAUSE DATA LOSS if not used very carefully.

Migrate of multiple containers to same S3 bucket with custom prefix
defined will cause over-write of matching object names.

[#155407966]


mdegerne:
Rebased on current master.

s3_sync/sync_s3.py
```diff
@@ -562,23 +566,51 @@ def _upload_part_worker(self, upload_id, s3_key, req_headers, queue,
...
         md5_hash = hashlib.md5('%s/%s' % (
             self.account.encode('utf-8'),
             self.container.encode('utf-8'))).hexdigest()
         # strip off 0x and L
         return hex(long(md5_hash, 16) % self.PREFIX_SPACE)[2:-1]
 
     def get_s3_name(self, key):
-        return u'%s/%s' % (self.get_prefix(), self._full_name(key))
+        prefix = self.get_prefix()
+        if self.use_custom_prefix:
```
dbishop: could this be
```
if self.use_custom_prefix:
    return '/'.join(filter(None, (prefix, key))
```

mdegerne:
Rebased on top of most recent master.

s3_sync/sync_s3.py
```diff
@@ -562,23 +563,36 @@ def _upload_part_worker(self, upload_id, s3_key, req_headers, queue,
...
+        return u'%s/%s' % (prefix, self._full_name(key))
 
     def get_manifest_name(self, s3_name):
-        # Split 3 times, as the format is:
-        # <prefix>/<account>/<container>/<object>
-        prefix, account, container, obj = s3_name.split('/', 3)
+        if self.use_custom_prefix:
+            prefix = self.get_prefix()
+            obj_prefix = '/'.join(filter(None, (prefix, '.manifests')))
+            obj = s3_name[len(prefix):].lstrip('/')
```
dbishop: This assignment to `obj` is dead code, right?

dbishop: Yup! my bad

PR #64, opened 2018-03-28
We should preserve the ETag for Swift uploads.

mdegerne:
When uploading objects into Swift, we used to replace the ETag header
with Content-MD5. As Swift ignores the Content-MD5, this is actually
wrong and we should rely on the ETag header.


mdegerne:
This was https://github.com/swiftstack/swift-s3-sync/pull/57, before I broke it. :(

PR #70, opened 2018-04-04
Run migrator only on demand

dbishop:
...in integration tests.

### DIFF
--- a/run_bash
+++ b/run_bash
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# This will be fast if your bouncy is already up to date
+docker pull bouncestorage/swift-aio
+# The --build is safe because if nothing in the Dockerfile has changed, it'll
+# just quickly "build" entirely from cache and not do anything.
+docker-compose up --build -d
+docker-compose exec -e DOCKER=true -e TERM=$TERM swift-s3-sync /bin/bash
--- a/test/integration/__init__.py
+++ b/test/integration/__init__.py
@@ -16,14 +16,66 @@ limitations under the License.
 
 import boto3
 import botocore.exceptions
+from contextlib import contextmanager
 import hashlib
 import json
 import os
+import psutil
+import signal
 import subprocess
 import swiftclient
 import time
 import unittest
 import urllib
+
+
+def wait_for_condition(timeout, checker):
+    start = time.time()
+    while time.time() < start + timeout:
+        ret = checker()
+        if ret:
+            return ret
+        time.sleep(0.1)
+    raise RuntimeError('Timeout (%s) expired' % timeout)
+
+
+def kill_a_pid(a_pid, timeout=4):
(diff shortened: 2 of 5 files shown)


test/integration/__init__.py
```diff
@@ -342,6 +384,31 @@ def nuser2_swift(self, method, *args, **kwargs):
     def s3(self, method, *args, **kwargs):
         return getattr(self.s3_client, method)(*args, **kwargs)
 
+    @contextmanager
+    def migrator_running(self):
```
timuralp: If we're going down this route, it might be cleaner to call the `s3_sync.migrator.run()` directly, with the appropriate arguments?

run_bash
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# This will be fast if your bouncy is already up to date
```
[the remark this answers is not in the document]
dbishop: Good point! I like silly things, but maybe this isn't the place for that.  I'll fix that.

test/integration/__init__.py
dbishop: I think this ends up exercising more code, and is pretty simple to reason about (run the daemon as it does in real life, and within X time, it should have succeeded at whatever the test is interested in)

dbishop:
I'll include the rationale in the commit message, but spoiler until I get to it: a bug can (and currently is) prevent the migrator from actually migrating, and the shunt makes it "feel" successful.  That's why having a Connection() that bypasses the shunt is necessary test tooling.  You can also have the shunt fail, but it's hard to test that if the migrator is always running and may have migrated the object(s), preventing the shunt from having anything to do.  We also have (had?) that bug.  So the tests need more control over when the background daemon performs migrations.

So the tests have more tools available to them now, all useful, and ready to be used more, to test more things and help us have more confidence.

timuralp:
Agreed that to exercise the shunt, the migrator can be problematic.

You can, however, examine whether objects have been migrated. When doing a GET on the container, the listing includes the object's location, so one can examine that to tell whether an object was moved or not. That's how the archive tests check whether an object was archived or not. Once the shunt was added for migrations, the migration tests should've been changed to examine the same. The relevant field in the listing is "content_location" when doing a JSON-format listing.

This is how archive tests do this:
```
 75         def _check_expired():
 76             # wait for the shunt to return the results for the object
 77             hdrs, listing = self.local_swift(
 78                 'get_container', mapping['container'])
 79             if int(hdrs['x-container-object-count']) != 0:
 80                 return False
 81             if any(map(lambda entry: 'content_location' not in entry,
 82                        listing)):
 83                 return False
 84             return (hdrs, listing)
```

dbishop:
I think @mdegerne will need this sooner than later, so I'll fix the bouncy thing in a sep. commit (like the bugfix I'm currently working on)

PR #71, opened 2018-04-04
Codify all containers swift-to-swift sync behavior

dbishop:
...in an integration test.

User newer container-crawler.

### DIFF
--- a/test/container/swift-s3-sync.conf
+++ b/test/container/swift-s3-sync.conf
@@ -141,13 +141,24 @@
             "aws_endpoint": "http://localhost:8080/auth/v1.0",
             "aws_identity": "\u062aacct2:\u062auser2",
             "aws_secret": "\u062apass2",
-            "container": "swift-restore",
             "container": "\u062acl-conn-swift-restore",
             "copy_after": 3600,
             "propagate_delete": false,
             "protocol": "swift",
             "retain_local": false,
             "restore_object": true
+        },
+        {
+            "account": "AUTH_slashacct",
+            "aws_bucket": "crazy-target:",
+            "aws_endpoint": "http://localhost:8080/auth/v1.0",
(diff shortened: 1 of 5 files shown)


test/container/swift-s3-sync.conf
```diff
@@ -141,13 +141,24 @@
...
             "container": "\u062acl-conn-swift-restore",
             "copy_after": 3600,
             "propagate_delete": false,
             "protocol": "swift",
             "retain_local": false,
             "restore_object": true
+        },
+        {
+            "account": "AUTH_slashacct",
+            "aws_bucket": "crazy-target",
```
timuralp: Can you change the bucket name to end with "/" or "-" so it will be easier to read the test output?

test/container/Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apt-get update && \
 
 RUN pip install -e git://github.com/swiftstack/botocore.git@1.4.32.6#egg=botocore && \
     pip install boto3==1.3.1 && \
-    pip install -e git://github.com/swiftstack/container-crawler.git@0.0.12#egg=container-crawler && \
+    pip install -e git://github.com/swiftstack/container-crawler.git@0.0.13#egg=container-crawler && \
```
timuralp: Would prefer this to be a separate commit.

test/container/swift-s3-sync.conf
dbishop: Pretty sure `/` is not allowed inside a container name by Swift, but a dash could be used.  I actually did that on purpose so it'd be obvious what happens if you just stick a normal string in there.

test/container/Dockerfile
dbishop: Really?

dbishop:
@timuralp updated w/suggestions

PR #73, opened 2018-04-05
Remove special-case for ProxyFS checkpoint container(s)

dbishop:


dbishop:
retest recheck

dbishop:
well, that's another way to spell "recheck" :(

dbishop:
Nice! So you can "redo" the check in TravisCI UI and it shows up as green, if it passes, here?

PR #75, opened 2018-04-05
Shunt and migrator fixes for alternate destination container name in Migrations

mdegerne:
The migrator was looking for headers based on the destination_container from
    the source provider.
    
    [Fixes #156481586]
    [Fixes #156480264]

    Add container GET to shunt migration behavior
    
    If the container does not exist locally, display the remote container
    listing instead (if available).
    
    Also fixed up migration tests for specifying a different container and for
    metadata tests to make use of some of the new test infrastructure.
    
    [Fixes #156480670]

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -50,7 +50,7 @@ IGNORE_KEYS = set(('status', 'aws_secret', 'all_buckets'))
 
 MigrateObjectWork = namedtuple('MigrateObjectWork', 'aws_bucket container key')
 UploadObjectWork = namedtuple('UploadObjectWork', 'container key object '
-                              'headers')
+                              'headers aws_bucket')
 
 
 class MigrationError(Exception):
@@ -310,14 +310,15 @@ class Migrator(object):
             q.put(None)
         q.join()
 
-    def _create_container(self, container, internal_client, timeout=1):
+    def _create_container(self, container, internal_client, aws_bucket,
+                          timeout=1):
         if self.config.get('protocol') == 'swift':
-            resp = self.provider.head_bucket(container)
+            resp = self.provider.head_bucket(aws_bucket)
             if resp.status == 404:
-                raise ContainerNotFound(container)
+                raise ContainerNotFound(aws_bucket)
             if resp.status != 200:
                 raise MigrationError('Failed to HEAD bucket/container "%s"' %
-                                     container)
+                                     aws_bucket)
             headers = {}
             propagated_hdrs = ['x-container-read', 'x-container-write']
             for hdr in resp.headers:
@@ -442,9 +443,9 @@ class Migrator(object):
         # x-history-location is configured), we should migrate the versions
         # before migrating the container itself.
         if self.config.get('protocol') == 'swift':
-            resp = self.provider.head_bucket(container)
+            resp = self.provider.head_bucket(aws_bucket)
             if resp.status == 404:
-                raise ContainerNotFound(container)
+                raise ContainerNotFound(aws_bucket)
             if resp.status != 200:
                 raise MigrationError('Failed to HEAD bucket/container "%s"' %
                                      container)
@@ -480,7 +481,7 @@ class Migrator(object):
             if not local:
                 if not ic.container_exists(self.config['account'],
                                            container):
-                    self._create_container(container, ic)
+                    self._create_container(container, ic, aws_bucket)
 
             remote = next(source_iter, None)
             if remote:
@@ -545,8 +546,10 @@ class Migrator(object):
             resp.body.close()
             self._migrate_slo(aws_bucket, container, key, put_headers)
         else:
-            self._upload_object(
-                container, key, FileLikeIter(resp.body), put_headers)
+            work = UploadObjectWork(
+                container, key, FileLikeIter(resp.body), put_headers,
+                aws_bucket)
+            self._upload_object(work)
 
     def _migrate_dlo(self, aws_bucket, container, key, headers):
         dlo_container, prefix = headers['x-object-manifest'].split('/', 1)
@@ -600,10 +603,12 @@ class Migrator(object):
         manifest_blob = json.dumps(manifest)
         headers['Content-Length'] = str(len(manifest_blob))
         work = UploadObjectWork(slo_container, key,
-                                FileLikeIter(manifest_blob), headers)
+                                FileLikeIter(manifest_blob), headers,
+                                slo_container)
         self.object_queue.put(work)
 
-    def _upload_object(self, container, key, content, headers):
+    def _upload_object(self, work):
+        container, key, content, headers, aws_bucket = work
         if 'x-timestamp' in headers:
             headers['x-timestamp'] = Timestamp(
                 float(headers['x-timestamp'])).internal
@@ -625,7 +630,7 @@ class Migrator(object):
             except UnexpectedResponse as e:
                 if e.resp.status_int != 404:
                     raise
-                self._create_container(container, ic)
+                self._create_container(container, ic, aws_bucket)
                 ic.upload_object(
                     content, self.config['account'], container, key,
                     headers)
@@ -643,7 +648,7 @@ class Migrator(object):
                     key = work.key
                     self._migrate_object(aws_bucket, container, key)
                 else:
-                    self._upload_object(*work)
+                    self._upload_object(work)
             except Exception:
                 self.errors.put((container, key, sys.exc_info()))
             finally:
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -16,6 +16,7 @@ limitations under the License.
 
 import json
 from lxml import etree
+import swiftclient
 
 from swift.common import constraints, swob, utils
 try:
@@ -203,19 +204,45 @@ class S3SyncShunt(object):
         # client-expected response.
         req.params = dict(req.params, format='json')
         status, headers, app_iter = req.call_application(self.app)
+
+        internal_resp = None
+        remote_iter = None
+        orig_status = None
+
+        if sync_profile.get('migration') and status.startswith('404 '):
+            # No local container, send remote only
+            try:
+                remote_iter = self.iter_remote(sync_profile, per_account,
+                                               marker, limit, prefix,
+                                               delimiter)
+            except swiftclient.exceptions.ClientException:
+                start_response(status, headers)
+                return app_iter
+
+            internal_resp = []
+            orig_status = status
+            status = '200 OK'
+
         if not status.startswith('200 '):
             # Only splice 200 (since it's JSON, we know there won't be a 204)
             start_response(status, headers)
             return app_iter
 
-        remote_iter = self.iter_remote(sync_profile, per_account, marker,
-                                       limit, prefix, delimiter)
+        if remote_iter is None:
+            remote_iter = self.iter_remote(sync_profile, per_account, marker,
+                                           limit, prefix, delimiter)
+
         remote_item, remote_key = next(remote_iter)
         if not remote_item:
-            start_response(status, headers)
+            if orig_status is not None:
+                start_response(orig_status, headers)
+            else:
+                start_response(status, headers)
             return app_iter
 
-        internal_resp = json.load(utils.FileLikeIter(app_iter))
+        if internal_resp is None:
+            internal_resp = json.load(utils.FileLikeIter(app_iter))
+
         spliced_response = []
         for local_item in internal_resp:
             if len(spliced_response) == limit:
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -221,13 +221,9 @@ class S3SyncShunt(object):
             if rem_resp.status == 200:
                 status = '200 OK'
                 headers = {}
-                propagated_hdrs = ['x-container-read', 'x-container-write',
-                                   'x-history-location', 'x-versions-location']
                 for hdr in rem_resp.headers:
-                    if hdr.startswith('x-container-meta-') or\
-                            hdr in propagated_hdrs:
-                        headers[hdr.encode('utf8')] = \
-                            rem_resp.headers[hdr].encode('utf8')
+                    headers[hdr.encode('utf8')] = \
+                        rem_resp.headers[hdr].encode('utf8')
                 internal_resp = []
 
         if not status.startswith('200 '):
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -16,7 +16,6 @@ limitations under the License.
 
 import json
 from lxml import etree
-import swiftclient
 
 from swift.common import constraints, swob, utils
 try:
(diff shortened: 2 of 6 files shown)


mdegerne:
Merged latest

dbishop:
Doesn't require change for this commit, but I think if we're going to stick things in a namedtuple, we should just use the namedtuple when we have one instead of almost immediately just yanking the contents back into variables.

So at some point, it'd be good to make `migrator.py` consistently use the named tuples when they're in-hand.

mdegerne:
Agreed about the named tuples. I'll try to fix next time I touch it the code.

test/integration/__init__.py
```diff
@@ -308,27 +308,46 @@ def setUpClass(klass):
...
             if mapping.get('aws_bucket') == '/*':
                 continue
-            if mapping['protocol'] == 'swift' and mapping.get('aws_bucket'):
-                # For now, the aws_bucket is just a prefix, not a container
-                # name for a swift destination that has a source container of
-                # /*.  So don't create a container of that name.
-                if mapping.get('container') != '/*':
-                    conn = klass.conn_for_acct_noshunt(mapping['aws_account'])
-                    conn.put_container(mapping['aws_bucket'])
+            if not container['aws_bucket'].startswith('no-auto-'):
```
dbishop: This might be easier to read as
```
if A:
    B
else:
    C
```
instead of
```
if not A:
    C
else:
    B
```

test/integration/test_migrator.py
```diff
@@ -37,13 +37,32 @@ def tearDown(self):
                 clear_swift_container(conn, container['container'])
                 clear_swift_container(conn,
                                       container['container'] + '_segments')
+                if container['container'].startswith('no-auto-'):
```
dbishop: It'd be nice to refactor this logic of "clear and delete a container in an account", so the logic was in one place and there were just 2 callers of it here.

test/integration/test_migrator.py
```diff
@@ -300,18 +319,35 @@ def test_migrate_new_container_location(self):
...
+                hdrs, listing = conn.get_container(migration['container'])
+                swift_names = [obj['name'] for obj in listing]
+                return test_object_set == set(swift_names)
+            except swiftclient.exceptions.ClientException as e:
+                if e.http_status == 404:
+                    return False
+                raise
 
         for name, body, headers in test_objects:
             self.remote_swift('put_object', migration['aws_bucket'], name,
```
dbishop: Another "doesn't need to change here", but I'm planning on getting rid of `self.remote_swift()` and friends, as I think they do more harm than good for test comprehensibility.

This line (which your diff doesn't modify, so this is no statement on you or your PR) is no more clear (I'd argue less clear given the frangibility of what "remote" can mean) than
```
conn_remote = self.conn_for_acct(migration['aws_account'])  # self-documenting target var name
...
conn_remote.put_object(migration['aws_bucket'], name, ...)
```
Note that the hard part of converting aws_* auth creds in the migration definition into an account name was already done in the base class's setUpClass (added in one of my patches).

The `self.remote_swift(...)` ONLY works for migration definitions that happen to have `aws_*` creds that map to `u'AUTH_\u062aacct2'`, which is unhelpfully limiting.

Anyway, this whole comment is just FYI and philosophy.

test/integration/test_migrator.py
```diff
@@ -300,18 +319,35 @@ def test_migrate_new_container_location(self):
...
-            wait_for_condition(5, _check_objects_copied)
+        try:
+            wait_for_condition(5, partial(_check_objects_copied, conn_local))
+        except RuntimeError as e:
+            if 'Timeout' in e.message:
+                hdrs, listing = conn_local.get_container(
+                    migration['container'])
+                swift_names = [obj['name'] for obj in listing]
+                self.assertEqual(test_object_set, set(swift_names))
+            else:
```
dbishop: Having to write this `else: ... raise` is a good argument for creating a `WaitTimedOut` exception that inherits from `RuntimeError` instead of using `RuntimeError` directly (in `wait_for_condition()` and propagated through the rest of the integration test codebase)

And fixing things like that is why my patches end up large... but who's going to just sit down to make that little bit better all by itself?

[Edited: but may be moot, here at least, since I don't think you need the wait_for_condition() at all]

test/integration/test_migrator.py
```diff
@@ -300,18 +319,35 @@ def test_migrate_new_container_location(self):
...
+                    return False
+                raise
 
         for name, body, headers in test_objects:
             self.remote_swift('put_object', migration['aws_bucket'], name,
                               StringIO.StringIO(body), headers=headers)
 
-        with self.migrator_running():
-            wait_for_condition(5, _check_objects_copied)
+        try:
```
dbishop: This is just a check that the shunt works, right?  If so, then just
```
self.assertTrue(_check_objects_copied(conn_local))
```
You don't need to wait on anything.

test/integration/test_migrator.py
```diff
@@ -300,18 +319,35 @@ def test_migrate_new_container_location(self):
...
+        except RuntimeError as e:
+            if 'Timeout' in e.message:
+                hdrs, listing = conn_local.get_container(
+                    migration['container'])
+                swift_names = [obj['name'] for obj in listing]
+                self.assertEqual(test_object_set, set(swift_names))
+            else:
+                raise
 
         for name, expected_body, user_meta in test_objects:
```
dbishop: If you spell this in terms of `conn` and put it in `_check_objects_copied()`, then you can have pre-migrator calls of
```
self.assertTrue(_check_objects_copied(conn_local))
self.assertTrue(_check_objects_copied(conn_remote))
# In fact, you can super-easily show the objects have NOT been migrated yet by this one line:
self.assertFalse(_check_objects_copied(conn_noshunt))
...
# then later, after wait_for_condition() has told you the migrator ran...
self.assertTrue(_check_objects_copied(conn_noshunt))
```

By doing this, you don't have two copies of this `for name, expected_body, user_meta in test_objects: ...` stuff.

dbishop:
Sorry, GitHub's formatting of my comments may not be very legible... you may want to go over the diff w/me in person.  If so just ask!

mdegerne:
Implemented changes suggested and rebased.

mdegerne:
Just identified a bug that I papered over in the test metadata stuff. (Thanks to @timuralp). Please do not merge yet.

mdegerne:
Added headers from remote container when migration has not yet moved the container. Note: I'll add separate story for supporting HEAD and for merging headers. The latter may require some input from product.

s3_sync/shunt.py
```diff
@@ -235,9 +241,14 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
             if orig_status is not None:
-                start_response(orig_status, headers)
-            else:
-                start_response(status, headers)
+                start_response(orig_status, orig_headers)
+                dict_headers = dict(headers)
+                res = self._format_listing_response([], resp_type, cont)
+                dict_headers['Content-Length'] = len(res)
+                dict_headers['Content-Type'] = resp_type
+                start_response(status, dict_headers.items())
```
dbishop: Can you call start_response twice?

mdegerne:
Resolved issue of lying (200 when s.b. 404).

test/integration/__init__.py
```diff
@@ -29,14 +29,18 @@
 import urllib
 
 
+class WaitTimedOut(RuntimeError):
```
dbishop: YES!

s3_sync/shunt.py
```diff
@@ -203,19 +204,45 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+        remote_iter = None
+        orig_status = None
+
+        if sync_profile.get('migration') and status.startswith('404 '):
+            # No local container, send remote only
+            try:
+                remote_iter = self.iter_remote(sync_profile, per_account,
+                                               marker, limit, prefix,
+                                               delimiter)
+            except swiftclient.exceptions.ClientException:
```
timuralp: This is unnecessary. Consider what `list_objects` does:
```
316     def list_objects(self, marker, limit, prefix, delimiter=None,
317                      bucket=None):
318         if bucket is None:
319             bucket = self.remote_container
320         try:
321             with self.client_pool.get_client() as swift_client:
322                 hdrs, results = swift_client.get_container(
323                     bucket, marker=marker, limit=limit,
324                     prefix=prefix, delimiter=delimiter)
325         except swiftclient.exceptions.ClientException as e:
326             return (e.http_status, e.message)
```

The other problem here is that what if we call this with boto in the future? We shouldn't break the abstraction of what specific client is being used behind the scenes.

s3_sync/migrator.py
```diff
@@ -210,6 +210,33 @@ def prune(self, migrations):
         self.save_status_list()
 
 
+def propagated_swift_headers(resp):
```
timuralp: Consider putting this into the utils module.

s3_sync/shunt.py
mdegerne: Agreed that I'm breaking the abstraction, but it is throwing ClientException at me now when the remote container does not exist. Alternative is catch a bare Exception. Is that what I want to do or should I find where the providers are throwing the exception(s) and change it there? Bare exceptions are Icky.

s3_sync/shunt.py
```diff
@@ -219,9 +220,14 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
 
-            internal_resp = []
             orig_status = status
-            status = '200 OK'
+            provider = create_provider(sync_profile, max_conns=1,
+                                       per_account=per_account)
+            rem_resp = provider.head_bucket(sync_profile['aws_bucket'])
+            if rem_resp.status == 200:
+                status = '200 OK'
+                headers = propagated_swift_headers(rem_resp)
```
timuralp: This doesn't seem right to me. We're going to set SYSMETA headers for versioning here? Are you sure that's what you're intending to do? Do we need to mutate the headers because of unicode handling in WSGI? What is the issue when we try to return the headers without munging them?

s3_sync/shunt.py
timuralp: Given the code sample above, I'm wondering where the ClientException is coming from. Have you figured out which specific call causes it? Looking through the code, I'm not sure how the exception makes it out of list_objects?

mdegerne:
I can't remember specifics right this moment, but un-munged headers broke
some tests, iirc. My reasoning for grabbing the whole section rather than
just the x-meta and propagated-header list was very shallow, mostly not
wanting to duplicate code. I'm easily convinced on this one.

s3_sync/shunt.py
dbishop: I can't speak to what this patch is doing, but if you're referring to this code:
```
+            # Dunno why, really, but the internal client app will 503
+            # with utf8-encoded header values IF the header key is a
+            # unicode instance (even if it's just low ascii chars in
+            # that unicode instance).  Go figure...
+            headers[hdr.encode('utf8')] = \
+                resp.headers[hdr].encode('utf8')
```
It existed before this patch, and the comment explains who had the problem (InternalClient app), and what caused it, even if it didn't make sense.  It fell out of exercising the code with more Unicode stuff.

That code needs to be where it was, but I dunno if it needs to also be in the new place.

timuralp: I'm commenting on using this code in the new place where InternalClient is *not* used.

s3_sync/shunt.py
mdegerne: I certainly can't tickle it out now, I wonder if this is a remnant from pre-refactor that I should have removed. I'll fix.

mdegerne:
Removed the weird munging of headers and resorted to somewhat normal munging of headers. Just a note @timuralp, using `rem_resp.headers` directly failed as did failure to do the utf8 encoding.

s3_sync/shunt.py
```diff
@@ -211,17 +210,25 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
-            internal_resp = []
+            remote_iter = self.iter_remote(sync_profile, per_account,
+                                           marker, limit, prefix,
+                                           delimiter)
+
             orig_status = status
-            status = '200 OK'
+            provider = create_provider(sync_profile, max_conns=1,
+                                       per_account=per_account)
+            rem_resp = provider.head_bucket(sync_profile['aws_bucket'])
```
timuralp: I suspect we can remove this extra HEAD request. Currently, list_objects tosses out the headers it gets back. What would it take to change it to return the headers in the tuple? We would need to do this both in the `sync_swift` and `sync_s3` `list_objects()` methods.

s3_sync/shunt.py
```diff
@@ -211,17 +210,25 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+            if rem_resp.status == 200:
+                status = '200 OK'
+                headers = {}
+                propagated_hdrs = ['x-container-read', 'x-container-write',
+                                   'x-history-location', 'x-versions-location']
```
timuralp: This doesn't seem enough to me. There could be `x-static-large-object` or `x-object-manifest` or `x-timestamp` and probably others. Looks like you've found that we have to call `encode(utf8)` on every header -- that makes sense. Are there any headers we would not want to propagate, though? Can we UTF-8 encode every header and avoid the white list? Otherwise, we definitely need to expand the white list.

s3_sync/shunt.py
```diff
@@ -203,19 +203,57 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+        if remote_iter is None:
+            remote_iter = self.iter_remote(sync_profile, per_account, marker,
+                                           limit, prefix, delimiter)
+
         remote_item, remote_key = next(remote_iter)
         if not remote_item:
+            if orig_status is not None:
+                dict_headers = dict(headers)
+                res = self._format_listing_response([], resp_type, cont)
+                dict_headers['Content-Length'] = len(res)
```
dbishop: I know the `swift.common.swob` wacky header class handles case-insensitivity, but are we running the risk of running afoul of it here?

The other places around here handling headers seem to be using all lowercase...

mdegerne:
Added a commit to pass all headers, not just those in white list.

mdegerne:
Fixed to REALLY pass all headers.

s3_sync/shunt.py
```diff
@@ -203,19 +203,53 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+                                           marker, limit, prefix,
+                                           delimiter)
+
+            orig_status = status
+            provider = create_provider(sync_profile, max_conns=1,
+                                       per_account=per_account)
+            rem_resp = provider.head_bucket(sync_profile['aws_bucket'])
+            if rem_resp.status == 200:
+                status = '200 OK'
+                headers = {}
```
timuralp: Could be 
```headers = dict([(k.encode('utf8'), v.encode('utf8')) for k, v in rem_resp.headers.items()])```

Don't have a preference. Arguably, the one-liner is more obtuse.

File: s3_sync/shunt.py [hunk 16]
```diff
@@ -203,19 +203,53 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+        if sync_profile.get('migration') and status.startswith('404 '):
+            # No local container, send remote only
+            remote_iter = self.iter_remote(sync_profile, per_account,
+                                           marker, limit, prefix,
+                                           delimiter)
+
+            orig_status = status
+            provider = create_provider(sync_profile, max_conns=1,
+                                       per_account=per_account)
+            rem_resp = provider.head_bucket(sync_profile['aws_bucket'])
```
timuralp: Let's merge this PR first, as this status munging will go away in the WSGI fix-up PR I have open.

s3_sync/shunt.py
```diff
@@ -203,19 +203,53 @@ def handle_listing(self, req, start_response, sync_profile, cont,
         # client-expected response.
         req.params = dict(req.params, format='json')
         status, headers, app_iter = req.call_application(self.app)
+
+        internal_resp = None
+        remote_iter = None
+        orig_status = None
+
+        if sync_profile.get('migration') and status.startswith('404 '):
```
timuralp: I think we can simplify things here. We always want to remove the remote store if it's a migration OR the container exists. Right? So, we could write something like:
```
if if not status.startswith('200 ') and not sync_profile.get('migration'):
    start_response(...)
    return app_iter

remote_iter = self.iter_remote(sync_profile, per_account, marker, limit, prefix, delimiter)
```

I think this would allow us to simplify some of the other logic around internal_resp and so on. Since I'll be removing the extra HEAD, I can probably clean this up.

timuralp: I guess that check would need to make sure that for migrations we got back a 404, so one more thing in the if-clause.

PR #76, opened 2018-04-05
Propagate DELETE container requests in migrations.

timuralp:
As we cannot set the X-Timestamp header when creating a container during
migrations, we have to make sure to propagate removal of containers to
the source cluster. As we don't do this currently, the containers will
be re-created on a subsequent migration run.

The commit also mutates the associated test to examine
the content_location information for the listings. This is necessary, as
after adding the support for migrations in the shunt we can no longer
simply examine the listing names.

### DIFF
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -419,13 +419,34 @@ class TestMigrator(TestCloudSyncBase):
         def _check_object_copied():
             hdrs, listing = self.local_swift(
                 'get_container', migration['container'])
-            names = [obj['name'] for obj in listing]
-            return names[0] == key
+            if not listing:
+                return False
+            if listing[0]['name'] != key:
+                return False
+            if 'swift' not in listing[0]['content_location']:
+                return False
+            return True
 
         def _check_removed():
             _, listing = self.local_swift(
                 'get_container', migration['container'])
             return listing == []
+
+        def _container_exists(client, container):
+            try:
+                client('get_container', container)
+                return True
+            except swiftclient.exceptions.ClientException as e:
+                if e.http_status == 404:
+                    return False
+                raise
+
+        def _check_removed_container():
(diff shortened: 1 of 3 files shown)


test/integration/test_migrator.py
```diff
@@ -419,14 +419,36 @@ def test_propagate_delete(self):
         def _check_object_copied():
             hdrs, listing = self.local_swift(
                 'get_container', migration['container'])
-            names = [obj['name'] for obj in listing]
-            return names[0] == key
+            if not listing:
+                return False
+            if listing[0]['name'] != key:
+                return False
+            if len(listing[0]['content_location']) == 1 or\
```
dbishop: Can this collapse to just
```
if 'swift' not in ...:
```
since there's no one-byte string that `swift` could be in?

test/integration/test_migrator.py
```diff
@@ -438,11 +460,13 @@ def _check_removed():
...
+            self.local_swift('delete_container', migration['container'])
+            wait_for_condition(5, _check_removed_container)
 
-        _, listing = self.local_swift('get_container', migration['container'])
-        self.assertEqual([], listing)
+        # recreate the removed container for future tests
+        self.local_swift('put_container', migration['container'])
+        self.remote_swift('put_container', migration['aws_bucket'])
 
         clear_swift_container(self.swift_dst, migration['aws_bucket'])
```
dbishop: Oh, I guess if the containers were deleted and PUT back, these two calls to `clear_swift_container()` can be removed.

timuralp: Done

test/integration/test_migrator.py
timuralp: `content_location` is a list, but you're right -- we don't have to explicitly check the length. Done.

PR #77, opened 2018-04-06
Migrations: send container metadata to the source.

timuralp:
During a migration, a container may not have been yet created on the
destination cluster. At this time, if a POST request is made, we should
propagate it to the source cluster. Returning a not found error in this
case would be confusing to the caller, as they would be able to GET and
HEAD the container, just not issue a POST.

### DIFF
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -467,6 +467,49 @@ class TestMigrator(TestCloudSyncBase):
         self.local_swift('put_container', migration['container'])
         self.remote_swift('put_container', migration['aws_bucket'])
 
+    def test_propagate_container_meta(self):
+        migration = self.swift_migration()
+        expected_meta = {
+            u'x-container-meta-migrated-\u062a': u'new-meta \u062a'}
+
+        # Since the shunt will propagate a container delete, we use the
+        # no-shunt proxy
+        conn_noshunt = self.conn_for_acct_noshunt(migration['account'])
+        conn_noshunt.delete_container(migration['container'])
+        self.local_swift('post_container', migration['container'],
+                         headers=expected_meta)
+
+        hdrs = self.remote_swift(
+            'head_container', migration['aws_bucket'])
+        self.assertEqual(
+            u'new-meta \u062a', hdrs.get(u'x-container-meta-migrated-\u062a'))
+
+        self.local_swift('put_container', migration['container'])
+        # The headers should no longer be visible on this container, as the
+        # local container should be authoritative and the shunt should not
+        # include any headers from the remote container.
+        new_hdrs = self.local_swift(
+            'head_container', migration['container'])
+        self.assertNotIn(u'x-container-meta-migrated-\u062a', new_hdrs)
+
+    def test_propagate_object_meta(self):
+        migration = self.swift_migration()
+        key = u'test_object-\u062a'
+        content = 'test object'
+        expected_meta = {
+            u'x-object-meta-migration-\u062a': u'new object meta \u062a'}
+
+        self.remote_swift('put_object', migration['aws_bucket'], key,
+                          StringIO.StringIO(content))
+        self.local_swift('post_object', migration['container'], key,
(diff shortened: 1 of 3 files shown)


test/integration/test_migrator.py
```diff
@@ -447,6 +447,40 @@ def _check_removed():
...
         clear_swift_container(self.swift_src, migration['container'])
 
+    def test_propagate_container_meta(self):
+        migration = self.swift_migration()
+        expected_meta = {'x-container-meta-migrated': 'new-meta'}
+
+        def _check_container_meta():
+            hdrs = self.remote_swift(
+                'head_container', migration['aws_bucket'])
+            return hdrs.get('x-container-meta-migrated') == 'new-meta'
```
dbishop: It doesn't look like the headers get fiddled with, but it'd be free to send a Unicode header key & value as well.

In other Unicode testing, I've found it most concise to leave the test input & such in UTF8 and then `.decode('utf8')` it back to a `unicode` instance when comparing with with comes back from a Swift API (for example).  Well, listing results were Unicode, but they came from JSON, so who knows.  The results here might be utf8-encoded.

In any case, adding Unicode input and expectation will improve the power of these tests.

test/integration/test_migrator.py
```diff
@@ -447,6 +447,40 @@ def _check_removed():
...
+
+        self.local_swift('delete_container', migration['container'])
+        self.local_swift('post_container', migration['container'],
+                         headers=expected_meta)
+        wait_for_condition(5, _check_container_meta)
+
+        self.local_swift('put_container', migration['container'])
```
dbishop: After this PUT to create the local container (without any extra headers), would you expect the container HEAD to have the headers or not?  In either case, putting an expectation in the test would be good for whichever it is.

And if the migrator running would change it, then that'd be good to add (but I don't know if it would).

test/integration/test_migrator.py
```diff
@@ -447,6 +447,40 @@ def _check_removed():
...
+
+    def test_propagate_object_meta(self):
+        migration = self.swift_migration()
+        key = 'test_object'
+        content = 'test object'
+        expected_meta = {'x-object-meta-migration': 'new object meta'}
```
dbishop: ditto on the Unicode biz.

Also, if migration of object names containing Unicode hasn't been tested elsewhere, you could cover that here, too.

test/integration/test_migrator.py
```diff
@@ -447,6 +447,40 @@ def _check_removed():
...
+        def _check_object_meta():
+            hdrs = self.remote_swift(
+                'head_object', migration['aws_bucket'], key)
+            return hdrs.get('x-object-meta-migration') == 'new object meta'
+
+        self.remote_swift('put_object', migration['aws_bucket'], key,
+                          StringIO.StringIO(content))
+        self.local_swift('post_object', migration['container'], key,
+                         headers=expected_meta)
+        wait_for_condition(5, _check_object_meta)
```
dbishop: FWIW, the Swift POST request should be synchronous w/the remote, right? Even if it weren't, there's no migrator running here, so there's nothing to wait on?  I.e. it could as easily be
```
self.assertTrue(_check_object_meta())
```

test/integration/test_migrator.py
timuralp: Done. This definitely did not make sense without asserting the header state. The migrator running right now would affect this, although there is an ask to propagate container metadata updates. I'll expand the test when that's finished.

test/integration/test_migrator.py
timuralp: I changed the test to use unicode values for testing. That definitely improves it -- done.

test/integration/test_migrator.py
timuralp: I was concerned about a situation where the a HEAD after POST may return stale results, but this is a single node test environment -- done.

test/integration/test_migrator.py
timuralp: Done.

PR #80, opened 2018-04-09
Handle unicode object metadata; remove the extra HEAD request; fix all-container migrations.

timuralp:
I found a few other bugs that the original PR uncovered. There are three total commits:
1. Properly handle all-container migrations
2. Removing the extra HEAD request (the refactoring is also important for the GET Account work)
3. Unicode fixes

It's probably best to review this one commit at a time. Collectively, the PR is a bit too big, but I tried to keep individual commits more manageable.

All-container migrations
--------------------------------

This commit fixes the case where we never change the `aws_bucket` parameter of the `sync_profile` to the container that the migration is currently attempting to process. This leads to 412 errors from the shunt (because `/*` is not a valid container name) and the shunt is de facto useless. Unfortunately, we did not have a test that covers this.

Removing the extra HEAD request
-----------------------------------------------

I ended up doing a bit of refactoring here. I think the logic is a bit simpler now, the extra HEAD is gone, and I managed to make `list_objects` use the same interface as the other similar methods. The patch changes the API to return a ProviderResponse (like the other
methods) and refactors the listing handling to take advantage of the
additional information. Majority of the changes are related to fixing
the tests for the API change.

Unicode fixes
------------------

This fell out when I tried to fix up the tests that attempted to use unicode in object names and headers. It's not entirely related to the other changes -- happy to split it out in a separate PR.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -372,21 +372,21 @@ class Migrator(object):
         next_marker = marker
 
         while True:
-            status, keys = self.provider.list_objects(
+            resp = self.provider.list_objects(
                 next_marker, self.work_chunk, prefix, bucket=container)
-            if status == 404:
+            if resp.status == 404:
                 raise ContainerNotFound(container)
-            if status != 200:
+            if resp.status != 200:
                 raise MigrationError(
                     'Failed to list source bucket/container "%s"' %
                     self.config['aws_bucket'])
-            if not keys and marker and marker == next_marker:
+            if not resp.body and marker and marker == next_marker:
                 raise StopIteration
-            for key in keys:
-                yield key
-            if not list_all or not keys:
+            for entry in resp.body:
+                yield entry
+            if not list_all or not resp.body:
                 break
-            next_marker = keys[-1]['name']
+            next_marker = resp.body[-1]['name']
         yield None
 
     def _check_large_objects(self, aws_bucket, container, key, client):
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -618,9 +618,6 @@ class Migrator(object):
             ts = Timestamp((ts - EPOCH).total_seconds()).internal
             headers['x-timestamp'] = ts
         del headers['last-modified']
-        # Encode all headers as UTF8
-        headers = {k.encode('utf8'): v.encode('utf8')
-                   for k, v in headers.items()}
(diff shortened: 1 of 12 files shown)


s3_sync/shunt.py
```diff
@@ -168,28 +207,34 @@ def __call__(self, env, start_response):
...
+                        internal_resp.status)
+                    break
+                if not internal_resp.body:
+                    break
+                for item in internal_resp.body:
+                    if 'name' in item:
+                        marker = item['name']
+                    else:
+                        marker = item['subdir']
+                    item['content_location'] = [item['content_location']]
```
dbishop: I know it's not new with this patch, but I'm just curious what made reassigning this as a one element array necessary?  Some WSGI requirement?

s3_sync/migrator.py
```diff
@@ -618,9 +618,6 @@ def _upload_object(self, work):
             ts = Timestamp((ts - EPOCH).total_seconds()).internal
             headers['x-timestamp'] = ts
         del headers['last-modified']
-        # Encode all headers as UTF8
```
dbishop: Is this call-site rendered unnecessary since the InternalClient runs the shunt code, and this code got moved to the shunt's `handle_object()` code path?

dbishop:
I'm going ahead and merging this since it's great and I need to get on with the release, but when you have a chance (not high prio at all), can you answer the curiosity questions?  Thanks!

s3_sync/shunt.py
timuralp: The only consumer of this today is the SwiftStack client, which examines the JSON listings and reports whether an object (or now container) is in the SwiftStack cluster, some public cloud, or both. To simplify the interface, the agreement is that the content_location is always going to be a list, with possible multiple entries. When we splice listings we take advantage of this by appending additional elements into the list (i.e. `swift` to indicate the SwiftStack cluster) if an element appears in both listings.

s3_sync/migrator.py
timuralp: Both the shunt and the migrator call the `utils.convert_to_local_headers`. I made that function handle the unicode encoding, so that we have one function that returns headers that can be passed to the internal client or used with WSGI: https://github.com/swiftstack/swift-s3-sync/blob/master/s3_sync/utils.py#L527

timuralp:
Thanks for merging this -- answered the questions on the review!

PR #82, opened 2018-04-09
Account GET (container listing) feature

timuralp:
It builds on the prior commits (separate PR for those). Focus on the last commit, which adds handling for GET on the account.

### DIFF
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -130,6 +130,63 @@ def _splice_listing(internal_resp, remote_iter, limit):
     return spliced_response
 
 
+def get_list_params(req, list_limit):
+    limit = int(req.params.get('limit', list_limit))
+    marker = req.params.get('marker', '')
+    prefix = req.params.get('prefix', '')
+    delimiter = req.params.get('delimiter', '')
+    path = req.params.get('path', None)
+    return limit, marker, prefix, delimiter, path
+
+
+def _format_xml_listing(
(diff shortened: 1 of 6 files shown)


s3_sync/shunt.py
```diff
@@ -170,14 +227,22 @@ def __init__(self, app, conf_file, conf):
...
 
         if not constraints.valid_api_version(vers):
             return self.app(env, start_response)
 
         if not cont:
+            sync_profile = self.sync_profiles.get((acct, '/*'))
+            if req.method == 'GET' and sync_profile and\
+                    sync_profile.get('migration'):
+                # TODO: make the container an optional parameter
+                profile, _ = maybe_munge_profile_for_all_containers(
```
dbishop: Agreed--use of `maybe_munge_profile_for_all_containers()` to get a provider that can't do anything to a specific container was not considered when originally written.  Optional arg makes sense, and it'll be the caller's responsibility (the code here) to not use that provider to not try to do something with the provider that requires a specific container name (which is obviously correct and good currently).

s3_sync/sync_swift.py
```diff
@@ -318,6 +328,15 @@ def _perform_op(client):
             with self.client_pool.get_client() as swift_client:
                 return _perform_op(swift_client)
 
+    def _make_content_location(self, bucket):
```
dbishop: +2 👍

PR #83, opened 2018-04-09
Shunt to work with HEAD container for migrations

mdegerne:
Specifically, when a container has not been migrated yet, the headers from
the source should be displayed.

### DIFF
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -261,14 +261,18 @@ class S3SyncShunt(object):
             return self.handle_delete(
                 req, start_response, sync_profile, obj, per_account)
 
-        if not obj and req.method == 'GET':
-            return self.handle_listing(req, start_response, sync_profile, cont,
-                                       per_account)
-        elif obj and req.method in ('GET', 'HEAD'):
+        if not obj:
+            if req.method == 'GET':
+                return self.handle_listing(
+                    req, start_response, sync_profile, cont, per_account)
+            if req.method == 'HEAD' and sync_profile.get('migration'):
+                return self.handle_container_head(
+                    req, start_response, sync_profile, cont, per_account)
+        if obj and req.method in ('GET', 'HEAD'):
             # TODO: think about what to do for POST, COPY
             return self.handle_object(req, start_response, sync_profile, obj,
                                       per_account)
-        elif req.method == 'POST':
+        if req.method == 'POST':
             return self.handle_post(req, start_response, sync_profile, obj,
                                     per_account)
 
@@ -368,6 +372,41 @@ class S3SyncShunt(object):
         dict_headers['Content-Type'] = resp_type
         start_response(status, dict_headers.items())
         return response
+
+    def handle_container_head(self, req, start_response, sync_profile, cont,
+                              per_account):
+        status, headers, app_iter = req.call_application(self.app)
+        if not status.startswith('404 '):
+            # Only shunt 404s
+            start_response(status, headers)
+            return app_iter
+
+        # Save off any existing trans-id headers so we can add them back later
+        trans_id_headers = [(h, v) for h, v in headers if h.lower() in (
+            'x-trans-id', 'x-openstack-request-id')]
+
+        provider = create_provider(sync_profile, max_conns=1,
+                                   per_account=per_account)
+
+        resp = provider.head_bucket(sync_profile['aws_bucket'])
+        if resp.status != 200:
+            # return original failure
+            start_response(status, headers)
+            return app_iter
+
+        headers = [(k.encode('utf-8'), unicode(v).encode('utf-8'))
+                   for k, v in resp.headers.iteritems()]
+        self.logger.debug('Remote resp: %s' % resp.status)
+
+        headers = filter_hop_by_hop_headers(headers)
+        headers.extend(trans_id_headers)
+
+        # TODO: Unfortunately, on HEAD bucket, S3 returns a 200 OK. To
+        # normalize this, we made the Swift provider also return 200 OK on
+        # HEAD. We should change the provider to return 204. When we fix the
+        # provider, we should return its return code.
+        start_response('204 No Content', headers)
+        return []
 
     def handle_object(self, req, start_response, sync_profile, obj,
                       per_account):
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -383,11 +383,20 @@ class TestMigrator(TestCloudSyncBase):
                     return False
                 raise
 
+        # verify get_head works through shunt
+        hdrs = conn_local.head_container(migration['container'])
+        self.assertIn('x-container-meta-test', hdrs)
+        self.assertEqual('test metadata', hdrs['x-container-meta-test'])
+
+        # Verify get_container works through shunt
         res = _check_container_created(conn_local)
         self.assertTrue(res)
         hdrs, listing = res
         self.assertIn('x-container-meta-test', hdrs)
         self.assertEqual('test metadata', hdrs['x-container-meta-test'])
+
+        # verify container not really there
+        self.assertFalse(_check_container_created(conn_noshunt))
 
         with self.migrator_running():
             hdrs, listing = wait_for_condition(5, partial(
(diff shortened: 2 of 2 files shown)


s3_sync/shunt.py
```diff
@@ -391,6 +395,40 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+        utils.close_if_possible(app_iter)
+
+        provider = create_provider(sync_profile, max_conns=1,
+                                   per_account=per_account)
+
+        resp = provider.head_bucket(sync_profile['aws_bucket'])
+        if resp.status != 200:
+            # return original failure
+            start_response(status, headers)
+            return app_iter
```
dbishop: You closed app_iter (if possible), above.  This is probably going to fail here?

s3_sync/shunt.py
```diff
@@ -391,6 +395,40 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+
+        self.logger.debug("MSD: remote-headers: %s" % resp.headers)
```
dbishop: Was this intended to be left in?

mdegerne:
Both valid complaints. Working....

s3_sync/shunt.py
```diff
@@ -391,6 +395,37 @@ def handle_listing(self, req, start_response, sync_profile, cont,
         start_response(status, dict_headers.items())
         return response
 
+    def handle_head(self, req, start_response, sync_profile, cont,
```
timuralp: I'd rename this to `container_head` to differentiate from object HEAD.

s3_sync/shunt.py
```diff
@@ -260,6 +260,10 @@ def __call__(self, env, start_response):
         if not obj and req.method == 'GET':
             return self.handle_listing(req, start_response, sync_profile, cont,
                                        per_account)
+        elif not obj and req.method == 'HEAD' and \
```
timuralp: Consider merging this with the prior if clause, which would be:
```
if not obj:
    if req.method == 'GET':
        ...
    elif req.method == 'HEAD':
        ...
```

This makes it obvious we're handling container requests.

test/integration/test_migrator.py
```diff
@@ -385,12 +385,21 @@ def _check_container_created(conn):
                     return False
                 raise
 
+        # verify get_head works through shunt
+        hdrs = conn_local.head_container(migration['container'])
+        self.assertIn('x-container-meta-test', hdrs)
+        self.assertEqual('test metadata', hdrs['x-container-meta-test'])
```
timuralp: Consider testing with unicode container metadata.

s3_sync/shunt.py
```diff
@@ -391,6 +395,37 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+
+        headers = [(k.encode('utf-8'), unicode(v).encode('utf-8'))
+                   for k, v in resp.headers.iteritems()]
+        self.logger.debug('Remote resp: %s' % status)
+
+        headers = filter_hop_by_hop_headers(headers)
+        headers.extend(trans_id_headers)
+
+        start_response('200 OK', headers)
+        return app_iter
```
timuralp: Same thing with `app_iter` that Darrell pointed out above.

s3_sync/shunt.py
```diff
@@ -391,6 +395,37 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+
+        resp = provider.head_bucket(sync_profile['aws_bucket'])
+        if resp.status != 200:
+            # return original failure
+            start_response(status, headers)
+            return app_iter
+
+        headers = [(k.encode('utf-8'), unicode(v).encode('utf-8'))
+                   for k, v in resp.headers.iteritems()]
+        self.logger.debug('Remote resp: %s' % status)
```
timuralp: Remote response is `resp.status`? I suspect copying/pasting code may be the issue.

s3_sync/shunt.py
```diff
@@ -391,6 +395,37 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+
+        headers = filter_hop_by_hop_headers(headers)
```
timuralp: We probably want to do it for all of our requests. I wonder if we can surface this to `__call__`.

mdegerne:
Fixed issues listed and rebased/merged commits. Note that the unicode test is the later test.

s3_sync/shunt.py
```diff
@@ -395,6 +399,37 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+
+        headers = [(k.encode('utf-8'), unicode(v).encode('utf-8'))
+                   for k, v in resp.headers.iteritems()]
+        self.logger.debug('Remote resp: %s' % resp.status)
+
+        headers = filter_hop_by_hop_headers(headers)
+        headers.extend(trans_id_headers)
+
+        start_response('200 OK', headers)
+        return app_iter
```
dbishop: Wait, is this returning `app_iter` which is from the local cluster, or the body of the response from the provider (remote cluster)?

s3_sync/shunt.py
```diff
@@ -395,6 +399,37 @@ def handle_listing(self, req, start_response, sync_profile, cont,
...
+            return app_iter
+
+        headers = [(k.encode('utf-8'), unicode(v).encode('utf-8'))
+                   for k, v in resp.headers.iteritems()]
+        self.logger.debug('Remote resp: %s' % resp.status)
+
+        headers = filter_hop_by_hop_headers(headers)
+        headers.extend(trans_id_headers)
+
+        start_response('200 OK', headers)
```
timuralp: 200 OK seems incorrect here (although, part of this is the fault of the provider API and S3). Swift returns 204 No Content on a container GET. S3 returns 200 OK. We should return 204 No Content on success.

s3_sync/shunt.py
timuralp: On success, this should be an empty body. Although, even on a 404, Swift seems to return an empty body, so we might be ok.

test/integration/test_migrator.py
```diff
@@ -444,17 +458,27 @@ def _check_objects_copied():
...
         for cont in test_containers:
-            self.swift_nuser.put_container(cont)
+            conn_remote.put_container(cont)
             for name, body, headers in test_objects:
-                self.nuser_swift('put_object', cont, name,
-                                 StringIO.StringIO(body), headers=headers)
+                conn_remote.put_object(
+                    cont, name, StringIO.StringIO(body), headers=headers)
+
+        hn = u'x-container-meta-tes\u00e9t'
```
timuralp: You don't have to address this, but I prefer somewhat clearer variable names.

s3_sync/shunt.py
dbishop: Oh, sorry, I forgot this was a HEAD handling function.  IIRC, the spec says HEAD bodies always have zero bytes (specifically, `...the server MUST NOT return a message-body in the response.`).

dbishop: ...that said, it seems like returning an empty body here would be more obviously correct (there's no _guarantee_ either of the other two responses correctly implemented HEAD and didn't include a message-body in their response...)

s3_sync/shunt.py
timuralp: Done

test/integration/test_migrator.py
timuralp: Done.

s3_sync/shunt.py
timuralp: Done -- returns [] now.

s3_sync/shunt.py
timuralp: Done.

s3_sync/shunt.py
timuralp: Done.

test/integration/test_migrator.py
timuralp: I changed this, as I think it does improve readability.

timuralp:
Ugh -- this is swiftclient screwing with us. It mutates the header dictionary that we pass it :(

Fix will be up shortly.

tipabu:
FWIW -- https://review.openstack.org/#/c/498069/ -- Stop mutating header dicts

PR #85, opened 2018-04-09
Use the proper ETag for SLO and content for DLO.

timuralp:
Our SLO/DLO uploads broke and we missed the 422 errors when attempting
to upload the manifests. There are two issues. For the SLO case, we
should be uploading the ETag of the *JSON content* rather than the ETag
returned by the origin Swift cluster (as there is no SLO middleware to
validate it).

For the DLO case, we have to GET the object with multipart-manifest=get
so that we retrieve the actual manifest blob. This is critical because a
DLO allows the manifest object to be non-zero size.

The patch fixes the large object tests that stopped working after the
migration shunt was put in place. There is also a small logger issue,
where we were no longer logging the actual requested objected on certain
failures.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -21,6 +21,7 @@ eventlet.patcher.monkey_patch(all=True)
 from collections import namedtuple
 import datetime
 import errno
+import hashlib
 import json
 import logging
 import os
@@ -534,6 +535,23 @@ class Migrator(object):
         args = {'bucket': aws_bucket}
         if self.config.get('protocol', 's3') == 'swift':
             args['resp_chunk_size'] = 65536
+
+        if (container, key) in self._manifests:
+            # Special handling for the DLO manifests
+            args['query_string'] = 'multipart-manifest=get'
+            resp = self.provider.get_object(key, **args)
+            if 'x-object-manifest' not in resp.headers:
+                self.logger.warning('DLO object changed before upload: %s/%s' %
+                                    (aws_bucket, key))
+                resp.body.close()
+                return
+            self._upload_object(UploadObjectWork(
+                container, key, FileLikeIter(resp.body),
+                convert_to_local_headers(resp.headers.items(),
+                                         remove_timestamp=False),
+                aws_bucket))
+            return
+
         resp = self.provider.get_object(key, **args)
         if resp.status != 200:
             resp.body.close()
@@ -541,11 +559,10 @@ class Migrator(object):
                 aws_bucket, key, resp.body))
         put_headers = convert_to_local_headers(
             resp.headers.items(), remove_timestamp=False)
-        if 'x-object-manifest' in resp.headers and\
-                (container, key) not in self._manifests:
+        if 'x-object-manifest' in resp.headers:
             self.logger.warning(
-                'Migrating Dynamic Large Object "%s/%s" -- results may not be '
-                'consistent' % (container, key))
+                'Migrating Dynamic Large Object "%s/%s" -- '
+                'results may not be consistent' % (container, key))
             resp.body.close()
             self._migrate_dlo(aws_bucket, container, key, put_headers)
         elif 'x-static-large-object' in resp.headers:
@@ -609,6 +626,10 @@ class Migrator(object):
             self.object_queue.put(work)
         manifest_blob = json.dumps(manifest)
         headers['Content-Length'] = str(len(manifest_blob))
+        # The SLO middleware is not in the pipeline. The ETag we provide should
+        # be for the manifest *JSON content*, rather than the hash of hashes
+        # that the SLO middleware can validate.
+        headers['etag'] = hashlib.md5(manifest_blob).hexdigest()
         work = UploadObjectWork(slo_container, key,
                                 FileLikeIter(manifest_blob), headers,
                                 slo_container)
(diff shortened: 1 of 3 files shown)


test/unit/test_migrator.py
```diff
@@ -1121,7 +1121,8 @@ def test_migrate_slo(self, create_provider_mock):
                     'x-object-meta-custom': 'slo-meta',
                     'x-timestamp': Timestamp(1.5e9).internal,
                     'x-static-large-object': 'True',
-                    'Content-Length': str(len(json.dumps(manifest)))}
+                    'Content-Length': str(len(json.dumps(manifest))),
+                    'etag': 'da843be9689d570dcd6f5f2ceaf9603f'}
```
dbishop: Can you add a comment above this line reminding future readers that the etag value is md5sum of the JSON manifest body (right?)?

s3_sync/migrator.py
```diff
@@ -651,15 +672,15 @@ def _upload_worker(self):
...
                 if isinstance(work, MigrateObjectWork):
-                    aws_bucket = work.aws_bucket
-                    container = work.container
-                    key = work.key
                     self._migrate_object(aws_bucket, container, key)
                 else:
                     self._upload_object(work)
             except Exception:
-                self.errors.put((container, key, sys.exc_info()))
+                self.errors.put((aws_bucket, key, sys.exc_info()))
```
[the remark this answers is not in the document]
timuralp: The `aws_bucket` is supposed to be the source of the object. This will give them an easy to find the object that is causing problems. The `container` is where it's being put. These two used to be always in sync, but Mandell's recent work allows the renaming of the destination container, so on failure to copy it, an error pointing to a non-existent object wouldn't help. There is a good argument for logging both things, though. I'd prefer not to redo the error checking stuff right now, though (to accommodate an extra value in the tuple).

test/unit/test_migrator.py
timuralp: You're right, this is sloppy. Changed to compute the etag from the json explicitly.

s3_sync/migrator.py
```diff
@@ -534,18 +535,34 @@ def _migrate_object(self, aws_bucket, container, key):
...
+            args['query_string'] = 'multipart-manifest=get'
+            resp = self.provider.get_object(key, **args)
+            if 'x-object-manifest' not in resp.headers:
+                self.logger.warning('DLO object changed before upload: %s/%s' %
+                                    (aws_bucket, key))
+                resp.body.close()
+                return
+            self._upload_object(UploadObjectWork(
+                container, key, FileLikeIter(resp.body),
+                convert_to_local_headers(resp.headers.items(),
```
tipabu: So does this pick up whatever ETag we got in the response?

timuralp: Yes. I think that's right here, because the `multipart-manifest=get` thing should return the object's ETag?

s3_sync/migrator.py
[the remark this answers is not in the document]
dbishop: 👍

s3_sync/migrator.py
dbishop: The body for that GET is empty-string (zero bytes) and teh ETag is teh md5sum of zero bytes in my testing.  Was that not what was expected?  Or is taht good?

dbishop: Specifically
```
root@a242b03f253a:/# curl -v -H 'X-Auth-Token: AUTH_tkdebe4e542f7849d6bbd8214e16abf1ba' http://localhost:8080/v1/AUTH_nacct/jojo/swift-s3-sync/LICENSE?multipart-manifest=get
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /v1/AUTH_nacct/jojo/swift-s3-sync/LICENSE?multipart-manifest=get HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:8080
> Accept: */*
> X-Auth-Token: AUTH_tkdebe4e542f7849d6bbd8214e16abf1ba
> 
< HTTP/1.1 200 OK
< Content-Length: 0
< Content-Type: application/octet-stream
< Accept-Ranges: bytes
< X-Object-Manifest: jojo_segments/swift-s3-sync/LICENSE/1517955763.000000/11357/5000/
< Last-Modified: Tue, 10 Apr 2018 21:39:28 GMT
< Etag: d41d8cd98f00b204e9800998ecf8427e
< X-Timestamp: 1523396367.00693
< X-Object-Meta-Mtime: 1517955763.000000
< X-Trans-Id: txce354032f061466c9c929-005acd2fef
< X-Openstack-Request-Id: txce354032f061466c9c929-005acd2fef
< Date: Tue, 10 Apr 2018 21:43:11 GMT
< 
* Connection #0 to host localhost left intact
root@a242b03f253a:/# md5sum ''
md5sum: : No such file or directory
root@a242b03f253a:/# md5sum< /dev/null
d41d8cd98f00b204e9800998ecf8427e  -
```

timuralp: @dbishop yep, that's how swiftclient does it. However, you can make a manifest object that also has content. I was testing that yesterday. Try it out! We should actually add an integration test (or mutate the current one) to check that -- I'll do that after updating one of Mandell's PRs.

dbishop: wild!

timuralp: @dbishop it gets wilder: you can make a manifest like so: `x-object-manifest: container/blobs-`, name the manifest `blobs` and put content in it. Then upload objects named `blobs-1` and `blobs-2`. When you do a GET for `blobs`, it will stitch together the content in `blobs`, `blobs-1`, and `blobs-2`. I was very confused that it's a thing that's allowed.

dbishop: wtf, that's crazy.  in any case, this patch looks great

PR #86, opened 2018-04-10
Bug/put on unmigrated container

mdegerne:
Move the migrator._create_container code to utils.

Intercept PUT object requests and create container when doesn't exist for migrations.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -33,11 +33,10 @@ from container_crawler.utils import create_internal_client
 from .daemon_utils import load_swift, setup_context, setup_logger
 from .provider_factory import create_provider
 from .utils import (convert_to_local_headers, convert_to_swift_headers,
-                    SWIFT_TIME_FMT, iter_listing)
+                    get_container_headers, iter_listing, RemoteHTTPError,
+                    SWIFT_TIME_FMT)
 from swift.common.internal_client import UnexpectedResponse
-from swift.common.middleware.versioned_writes import (
-    SYSMETA_VERSIONS_LOC, SYSMETA_VERSIONS_MODE)
-from swift.common.swob import Request
+from swift.common import swob
 from swift.common.utils import FileLikeIter, Timestamp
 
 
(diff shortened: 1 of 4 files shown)


s3_sync/utils.py
```diff
@@ -482,6 +499,63 @@ def __del__(self):
...
+
+    start = time.time()
+    while time.time() - start < timeout:
+        if not internal_client.container_exists(
+                migrate_config['account'], container):
+            time.sleep(0.1)
+        else:
+            logger.debug('Created container "%s"' % container)
+            return
+    raise MigrationError('Timeout while creating container "%s"' % container)
```
timuralp: This makes less sense in the utils class, especially if the method is used outside of the migrator.

s3_sync/migrator.py
```diff
@@ -53,19 +51,6 @@
                               'headers aws_bucket')
 
 
-class MigrationError(Exception):
```
timuralp: I think we should keep this exception here and amend the `create_container` code to not raise this exception, as it may not be called from the migrator anymore.

s3_sync/utils.py
```diff
@@ -482,6 +499,63 @@ def __del__(self):
         self.close()
 
 
+def migration_create_container(container, internal_client, aws_bucket,
+                               provider, migrate_config, logger, timeout=1):
```
timuralp: If we pass the provider, we shouldn't have to pass the configuration dictionary. You can check the provider class. The other place we use the config is to get the account where the container is created. I'd rather we pass the account explicitly for that.

s3_sync/utils.py
```diff
@@ -482,6 +499,63 @@ def __del__(self):
...
+    if migrate_config.get('protocol') == 'swift':
```
timuralp: This would be `if isinstance(provider, SyncSwift)` if we don't pass the configuration.

s3_sync/utils.py
```diff
@@ -482,6 +499,63 @@ def __del__(self):
         self.close()
 
 
+def migration_create_container(container, internal_client, aws_bucket,
```
timuralp: Maybe `create_internal_container`? This method could be used in general to create containers with the internal client.

s3_sync/utils.py
```diff
@@ -482,6 +499,63 @@ def __del__(self):
...
+                               provider, migrate_config, logger, timeout=1):
+    if migrate_config.get('protocol') == 'swift':
+        resp = provider.head_bucket(aws_bucket)
+        if resp.status == 404:
+            raise ContainerNotFound(aws_bucket)
+        if resp.status != 200:
+            raise MigrationError('Failed to HEAD bucket/container "%s"' %
```
timuralp: I would not raise a MigrationError here.

s3_sync/shunt.py
```diff
@@ -215,6 +220,13 @@ def __init__(self, app, conf_file, conf):
                    cont['container'].encode('utf-8'))
             self.sync_profiles[key] = cont
 
+        pool_size = 10  # TODO(MSD) make better choice
+        swift_dir = conf.get('swift_dir', '/etc/swift')
+        self.internal_pool = eventlet.pools.Pool(
```
timuralp: This is the wrong approach here. In the shunt context, we are already running in the proxy server. Notice how the `SwiftPutWrapper`, for example, does not create an internal client. We should be able to call `request.get_response(app)` to execute our container creation request.

This might actually mean that we can't reuse the entirety of the migrator's container creation code, but rather factor out the code that creates the request and only reuse that.

s3_sync/shunt.py
```diff
@@ -272,6 +284,10 @@ def __call__(self, env, start_response):
             return self.handle_post(req, start_response, sync_profile, obj,
                                     per_account)
 
+        if obj and req.method == 'PUT' and sync_profile.get('migration'):
```
timuralp: I'm now realizing that I should change the `handle_post` to split it up for container vs object, so we can structure these clauses as Account methods, Container methods, Object methods. I wouldn't change anything in the PR here, but hopefully I'll remember to do this soon! :)

mdegerne:
That was the same conclusion I had come to, which is why I was naively using internal client to check for the existence of the container.

PR #90, opened 2018-04-13
Fix unicode handling in S3 migrations.

timuralp:
Fixes handling of objects with unicode names and removes a test for
unicode characters in the metadata, as S3 does not support unicode
characters:
"Amazon S3 stores user-defined metadata keys in lowercase.
 Each key-value pair must conform to US-ASCII when you are using REST."
-- [1]

[1] https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

### DIFF
--- a/s3_sync/shunt.py
+++ b/s3_sync/shunt.py
@@ -479,6 +479,7 @@ class S3SyncShunt(object):
         if req.method == 'GET' and sync_profile.get('restore_object', False) \
                 and 'range' not in req.headers:
             # We incur an extra request hit by checking for a possible SLO.
+            obj = obj.decode('utf-8')
             manifest = provider.get_manifest(obj)
             self.logger.debug("Manifest: %s" % manifest)
             status, headers, app_iter = provider.shunt_object(req, obj)
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -68,10 +68,9 @@ class TestMigrator(TestCloudSyncBase):
 
         test_objects = [
             ('s3-blob', 's3 content', {}, {}),
-            ('s3-unicod\u00e9', '\xde\xad\xbe\xef', {}, {}),
+            (u's3-unicod\u00e9', '\xde\xad\xbe\xef', {}, {}),
             ('s3-with-headers', 'header-blob',
-             {'custom-header': 'value',
-              'unicod\u00e9': '\u262f'},
+             {'custom-header': 'value'},
              {'content-type': 'migrator/test',
               'content-disposition': "attachment; filename='test-blob.jpg'",
               'content-encoding': 'identity'})]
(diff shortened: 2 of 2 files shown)


tipabu:
So what do we do when syncing Swift objects with non-ASCII meta to S3? Even more fun, how about non-UTF-8 meta?

Even if we aren't testing non-ASCII, we should probably test non-printable ASCII. I've had to [make it work](https://github.com/openstack/swift3/blob/master/swift3/request.py#L926-L936) in swift3, and it's kind of weird -- don't trust everything you read in the docs, we've got to be empiricists ☹️

timuralp:
@dbishop @tipabu well, boto doesn't let us PUT an object with non-ascii characters. Here is the backtrace:
```
c.put_object(Bucket='gerard-test-bucket', Key='test-non-ascii', Metadata={'foo': u'bár'})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/timur/swiftstack/s3-sync/venv/src/botocore/botocore/client.py", line 278, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/timur/swiftstack/s3-sync/venv/src/botocore/botocore/client.py", line 548, in _make_api_call
    api_params, operation_model, context=request_context)
  File "/home/timur/swiftstack/s3-sync/venv/src/botocore/botocore/client.py", line 598, in _convert_to_request_dict
    params=api_params, model=operation_model, context=context)
  File "/home/timur/swiftstack/s3-sync/venv/src/botocore/botocore/hooks.py", line 227, in emit
    return self._emit(event_name, kwargs)
  File "/home/timur/swiftstack/s3-sync/venv/src/botocore/botocore/hooks.py", line 210, in _emit
    response = handler(**kwargs)
  File "/home/timur/swiftstack/s3-sync/venv/src/botocore/botocore/handlers.py", line 490, in validate_ascii_metadata
    report=error_msg)
```

If we remove the ASCII validator. The next failure we get is a signature mismatch. There is a boto3 thread about this: https://github.com/boto/boto3/issues/478. In it they mentioned that  I don't know if we need to go out of our way to fight it for migrations. If we want to pursue this, we would need to dig into the signer and figure out what the double encoding that they're referring to is.

I also tried setting it through the AWS S3 UI, which reported back "Key-value pair must conform to US-ASCII standard". `s3cmd` reports a similar error.

It might be interesting to see what happens when an S3 clone, such as swift3, allows this and what the library will do. I suspect this will be broken in botocore. That seems like a separate thing for us to chase down?

@tipabu I totally agree on putting things _into_ S3. I was thinking we should base64 encode the keys and values in that case. It's something that we likely have to do to support Azure anyway, so that has some additional utility. But this PR is about pulling objects _out_ of S3.

tipabu:
Still -- we should test with _non-printable_ ASCII! `'\x04'`, say, like [ceph/s3-tests uses](https://github.com/ceph/s3-tests/blob/master/s3tests/functional/test_s3.py#L1304-L1317).

timuralp:
Agreed on that -- will amend to include non-printable characters.

timuralp:
Well, I spent a lot of time today in this particular rabbit hole. Filed an issue against S3Proxy: https://github.com/gaul/s3proxy/issues/271. Because of that, I think we should track internally having a test and code support for non-printable characters during migrations.

PR #92, opened 2018-04-13
Gut cloud-connector GET/HEAD and add unit tests

dbishop:
Cloud connector app.py is now 100% covered by unit tests.

TODO: get full test coverage for cloud_connector/{auth,util}.py


test/unit/test_cloud_sync_app.py
```diff
@@ -0,0 +1,422 @@
...
+
+        def _config_writer(*args, **kwargs):
+            with open(self.cloud_connector_sync_conf_path, 'wb') as fh:
+                json.dump(self.sync_conf, fh)
+                fh.flush()
+
+        mock_get_and_write.side_effect = _config_writer
+
+        conf = {'swift_baseurl': self.swift_baseurl}
+        self.fake_logger = FakeLogger()
```
timuralp: We pull in `FakeLogger` and set it up, but don't actually use it. Currently, would anything not work if we used a mock for the logger? I'm still unclear on whether we want to use it or not and, I may have missed it, but it doesn't look like it's currently used?

timuralp:
@dbishop I didn't realize you were not going to amend the first commit. That looks silly now and I agree that you could've left the FakeLogger stuff in. My prior suggestion was to amend the commit to keep history cleaner.

dbishop:
*shrug*, FakeLogger will come back if it ends up getting used.  Removing it
was pretty easy.

PR #93, opened 2018-04-18
Add S3 API authen/authz for cloud-connector

dbishop:
We decided to only support the S3 API in the first pass of
cloud-connector.  So the auth had to be rewritten and swift3
incorporated.

cloud-connector now requires a DB mapping S3 API key IDs to secret key
values and Swift Account names.

There still aren't any integration tests because the cloud-connector's
controller is still a dumb stub that just returns NotImplemented() for
all verbs.


s3_sync/cloud_connector/auth.py
```diff
@@ -43,190 +38,88 @@ def __init__(self, app, conf):
...
-
-        Otherwise, the request requires an X-Auth-Token header and its value
-        must be in our cache.  Furthermore, for any specific token, we only
-        allow accesses for the account that was in the returned storage URL
-        from the onprem Swift cluster's auth response that delivered up that
-        token.
+        TODO(darrell): add support for Swift v1 authen/authz
+        TODO(darrell): add support for X-Auth-Token header
+
+        Currently only supports authen (via swift3 callback) and authz for S3
```
timuralp: Is "authen" a function/API name somewhere? I don't see us calling it, but just in comments?

PR #94, opened 2018-04-18
Make container last-modified entry optional.

timuralp:
When listing containers (i.e. GET <Account>), the last-modified entry
for each container may not exist. This is true in old Swift clusters and
luckily we don't have to rely on it.


timuralp:
This is required for snapfish. I'm going to build a package for them with this commit as version 0.1.30.1, branched from the last tag.

PR #99, opened 2018-04-25
Add provider put_object() methods.

dbishop:
The cloud-connector will need to be able to just PUT an object, to
either S3 or Swift, sanely (i.e. accepting an iterable of body, not
requiring a known length).

Also beefs up some of the "wrapper" helper classes.

All the new and changed code has 100% line and branch unit test
coverage, AFACT.

### DIFF
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -23,8 +23,7 @@ from swift.common.utils import FileLikeIter
 import traceback
 import urllib
 
-from .base_sync import BaseSync
-from .base_sync import ProviderResponse
+from .base_sync import BaseSync, ProviderResponse
 from .utils import (FileWrapper, ClosingResourceIterable, check_slo,
                     SWIFT_USER_META_PREFIX, SWIFT_TIME_FMT)
 
@@ -70,6 +69,11 @@ class SyncSwift(BaseSync):
     def _close_conn(conn):
         if conn.http_conn:
             conn.http_conn[1].request_session.close()
+
+    def put_object(self, swift_key, headers, body_iter, query_string=None):
+        return self._call_swiftclient('put_object', self.container, swift_key,
+                                      contents=body_iter, headers=headers,
+                                      query_string=query_string)
 
     def upload_object(self, swift_key, policy, internal_client):
         if self._per_account and not self.verified_container:
@@ -301,9 +305,13 @@ class SyncSwift(BaseSync):
                 else:
                     headers = resp
                     body = ['']
-                status = 206 if 'content-range' in headers else 200
-                headers = dict([translate(header, value)
-                                for header, value in headers.items()])
+                if 'response_dict' in args:
+                    headers = args['response_dict']['headers']
+                    status = args['response_dict']['status']
+                else:
+                    status = 206 if 'content-range' in headers else 200
+                    headers = dict([translate(header, value)
+                                    for header, value in headers.items()])
                 return ProviderResponse(True, status, headers, body)
             except swiftclient.exceptions.ClientException as e:
                 headers = dict([translate(header, value)
@@ -315,6 +323,7 @@ class SyncSwift(BaseSync):
                 self.logger.exception('Error contacting remote swift cluster')
                 return ProviderResponse(False, 502, {}, iter('Bad Gateway'))
 
+        # TODO: always use `response_dict` biz
         if op == 'get_object' and 'resp_chunk_size' in args:
             entry = self.client_pool.get_client()
             resp = _perform_op(entry.client)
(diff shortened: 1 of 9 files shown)


s3_sync/sync_swift.py
```diff
@@ -296,14 +300,20 @@ def _perform_op(client):
                 if not resp:
                     return ProviderResponse(True, 204, {}, [''])
 
+                body = None
```
timuralp: Since `body` can't be undefined after the following if/else, do we need this?

s3_sync/sync_swift.py
```diff
@@ -296,14 +300,20 @@ def _perform_op(client):
...
                 if isinstance(resp, tuple):
                     headers, body = resp
                 else:
                     headers = resp
                     body = ['']
-                status = 206 if 'content-range' in headers else 200
-                headers = dict([translate(header, value)
-                                for header, value in headers.items()])
+                if 'response_dict' in args:
```
timuralp: We're using `response_dict` as a proxy for checking if it's a `put_object` call. Maybe it would be better to explicitly check the op? I'm not too fond of the `tuple` check above that we're doing to differentiate the responses.

s3_sync/sync_swift.py
```diff
@@ -326,6 +336,9 @@ def _perform_op(client):
                     entry, resp.body, lambda: None)
             return resp
         else:
+            if op == 'put_object':
+                response_dict = args.get('response_dict', {})
```
timuralp: I wonder if we should do this for all requests. I don't want to make this PR bigger, but could you add a TODO here?

s3_sync/utils.py
```diff
@@ -73,61 +73,174 @@ def __unicode__(self):
...
+    def seek(self, pos, flag=0):
+        if pos != 0:
+            raise RuntimeError('%s: arbitrary seeks are not supported' %
+                               self.__class__.__name__)
+        if self._bytes_delivered == 0:
+            return
+        if self.seek_zero_cb:
+            self.iterator = self.seek_zero_cb()
+            self.buf = None
+        # Without a callback, we leave self.buf alone (there's no truly "right"
```
timuralp: Is there a use case where we want to allow resetting the stream without a callback? Maybe we should raise an exception instead, as otherwise the returned data may actually be corrupted and not at all what the caller expects.

s3_sync/utils.py
```diff
@@ -73,61 +73,174 @@ def __unicode__(self):
...
+            raise StopIteration
+        if self.buf:
+            rv = self.buf
+            self.buf = None
+        else:
+            rv = next(self.iterator)
+        if not called_from_read:
+            # If we *were* called from read(), then it will do the truncation,
+            # if necessary, and accounting.
+            rv = self._account_data_delivered(rv)
```
timuralp: Two questions:
1. Should we be raising `StopIteration` if we cannot return more data (`self.length == self._bytes_delivered`)?
2. I wonder if we need to be closing the underlying socket somehow when the length is less than the amount of data the iterator returns?

s3_sync/utils.py
```diff
@@ -73,61 +73,174 @@ def __unicode__(self):
...
+        if self._swift_stream:
+            self._swift_stream.close()
+
         status, self._headers, body = self._swift.get_object(
             self._account, self._container, self._key,
             headers=self.swift_req_hdrs)
         if status != 200:
             raise RuntimeError('Failed to get the object')
-        self._bytes_read = 0
+
```
timuralp: Spurious new line?

test/integration/test_s3_sync.py
```diff
@@ -168,21 +170,84 @@ def get_etag(key):
             self._test_archive(key, content, mapping, get_etag,
                                expected_location)
 
+    def test_provider_s3_put_object_defaults(self):
```
timuralp: I would collapse the three `put_object` tests, as most of the test case is repeated.

test/unit/test_sync_s3.py
```diff
@@ -51,6 +51,211 @@ def setUp(self, mock_boto3):
                                'container': 'container'},
                               max_conns=self.max_conns)
 
+    @mock.patch('s3_sync.sync_s3.SeekableFileLikeIter')
+    def test_put_object(self, mock_seekable):
```
timuralp: I suspect the put_object tests could also be collapsed, as most of it is duplicated and we could iterate over test cases. Fine as is.

timuralp:
@dbishop since you rebased this branch, probably could've squashed the two commits together. It's ok otherwise -- I'll probably squash them when merging.

s3_sync/utils.py
dbishop: will fix

s3_sync/utils.py
dbishop: (IRL) no changes needed

s3_sync/utils.py
dbishop: probably not

test/integration/test_s3_sync.py
dbishop: (IRL) it's fine

test/unit/test_sync_s3.py
dbishop: (IRL) it's fine

PR #100, opened 2018-04-25
Fix import statement of VERSIONS for old swifts

mdegerne:
And behave sanely if we don't have it.


timuralp:
@mdegerne oh, that's a little funny. I will approve my own commit then :)

PR #104, opened 2018-04-27
Add coverage for the daemon_utils module.

timuralp:
Adds tests for setting up the logger and daemon context.

### DIFF
--- a/test/unit/test_daemon_utils.py
+++ b/test/unit/test_daemon_utils.py
@@ -1,0 +1,96 @@
+import mock
+import s3_sync.daemon_utils
+import StringIO
+import sys
+import unittest
+
+
+class TestDaemonUtils(unittest.TestCase):
+
+    def setUp(self):
+        self.old_swift_module = sys.modules.get('swift')
+        if 'swift' in sys.modules:
+            del sys.modules['swift']
+
+    def tearDown(self):
+        if 'swift' in sys.modules:
+            del sys.modules['swift']
+        if self.old_swift_module:
+            sys.modules['swift'] = self.old_swift_module
+
+    @mock.patch('s3_sync.daemon_utils.logging')
+    def test_old_swift_raises_exception(self, logging_mock):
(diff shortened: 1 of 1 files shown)


test/unit/test_daemon_utils.py
```diff
@@ -0,0 +1,96 @@
...
+            del sys.modules['swift']
+
+    def tearDown(self):
+        if 'swift' in sys.modules:
+            del sys.modules['swift']
+        if self.old_swift_module:
+            sys.modules['swift'] = self.old_swift_module
+
+    @mock.patch('s3_sync.daemon_utils.logging')
+    def test_old_swift_raises_exception(self, loggin_mock):
```
dbishop: 2nd arg probably supposed to be `logging_mock`?

test/unit/test_daemon_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+        import_mock.side_effect = _import
+        with self.assertRaises(ImportError):
+            s3_sync.daemon_utils.load_swift('logger', once=True)
+        import_mock.assert_called_once_with('swift', mock.ANY, None, None)
+
+    @mock.patch('sys.stdout', new_callable=StringIO.StringIO)
+    @mock.patch('s3_sync.daemon_utils.sys')
+    @mock.patch('s3_sync.daemon_utils.os.path.exists')
+    def test_setup_context_exits(self, exists_mock, sys_mock, mock_stdout):
```
dbishop: I tend to prefer `mock_*` for mock var names, but as long as they're consistent w/in a small-ish scope, it doesn't matter.  But here, the 4th arg should probably be `stdout_mock` for consistency.

test/unit/test_daemon_utils.py
timuralp: Done

test/unit/test_daemon_utils.py
timuralp: Done

PR #106, opened 2018-04-30
Allow older_than (seconds) in migrations

mdegerne:
Will only migrate objects whose last_modification date is older than the
specified number of seconds. If not specified, it behaves as before.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -654,6 +654,16 @@ class Migrator(object):
         return self._find_missing_objects(container, aws_bucket, marker,
                                           prefix, list_all)
 
+    def _old_enough(self, remote):
+        older_than = self.config.get('older_than')
+        if older_than is None:
+            return True
+        older_than = datetime.timedelta(seconds=older_than)
+        now = datetime.datetime.utcnow()
+        remote_time = datetime.datetime.strptime(
+            remote['last_modified'], SWIFT_TIME_FMT)
+        return remote_time < now - older_than
+
     def _find_missing_objects(
             self, container, aws_bucket, marker, prefix, list_all):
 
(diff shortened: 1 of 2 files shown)


s3_sync/migrator.py
```diff
@@ -524,6 +524,18 @@ def _process_container(
         return self._find_missing_objects(container, aws_bucket, marker,
                                           prefix, list_all)
 
+    def _old_enough(self, remote):
+        if 'last_modified' not in remote:
```
timuralp: Hopefully, this never happens. I think a `KeyError` in that case is actually appropriate.

test/unit/test_migrator.py
```diff
@@ -693,6 +694,9 @@ def test_migrate_objects(self, create_provider_mock):
         provider = create_provider_mock.return_value
         self.migrator.status.get_migration.return_value = {}
 
+        utcnow = datetime.datetime.utcnow()
```
timuralp: I'd consider stubbing the `datetime.datetime.utcnow` call to test various scenarios in the unit test (and avoids having to deal with flooring the time).

s3_sync/migrator.py
mdegerne: TBH: This check exists to save having to modify a bunch of tests that didn't need to include last_modified in the data.

timuralp: That is a little confusing -- wouldn't those tests pass because `older_than` is not configured anyway and we should get out of this on line 532?

test/unit/test_migrator.py
mdegerne: I'm not convinced that mocking this is less effort than working with real time in the tests. Having the tests set times relative to current time is, imo, a clearer demonstration of what is being tested than being able to arbitrarily change the time.

s3_sync/migrator.py
mdegerne: You're right. I instituted this before I had the current order of operations. Deleting.

PR #107, opened 2018-05-01
Fix container date comparison

mdegerne:
Last modified is updated on POST where x-timestamp is not.


s3_sync/migrator.py
```diff
@@ -140,12 +140,12 @@ def _update_status_counts(status, moved_count, scanned_count, reset):
 
 
 def _create_x_timestamp_from_hdrs(hdrs):
-    if 'x-timestamp' in hdrs:
-        return float(hdrs['x-timestamp'])
     if 'last-modified' in hdrs:
         ts = datetime.datetime.strptime(hdrs['last-modified'],
                                         LAST_MODIFIED_FMT)
         return (ts - EPOCH).total_seconds()
+    if 'x-timestamp' in hdrs:
```
timuralp: Is there a case where `last-modified` is not present, but `x-timestamp` is? I would expect us to not want to look at `x-timestamp` at all.

s3_sync/migrator.py
```diff
@@ -511,11 +511,19 @@ def _process_container(
                     remote_ts = _create_x_timestamp_from_hdrs(resp.headers)
                     if local_ts is not None and remote_ts is not None and \
                             local_ts < remote_ts:
+                        if self.config['container'] == 'metadata_test':
```
timuralp: Remove these lines. Seems like it's left over from debugging.

s3_sync/migrator.py
```diff
@@ -511,11 +511,19 @@ def _process_container(
...
+                            self.logger.debug('MSD: updating container')
                         header_changes = diff_container_headers(
                             resp.headers, local_headers)
                         if header_changes:
                             self._update_container_headers(
                                 container, ic, header_changes)
+                    else:
+                        if self.config['container'] == 'metadata_test':
```
timuralp: Same comment as above about debug lines.

test/integration/test_migrator.py
```diff
@@ -658,6 +661,7 @@ def key_gone(key):
             wait_for_condition(5, key1_copied)
             self.assertTrue(key_gone(key2))
 
+            time.sleep(2)
```
timuralp: I wonder if we can squash some of the cases in this test, so that we don't sleep twice for 4 seconds. Specifically, we want to test that we 1. delete a header that's no longer present and 2. update the value of a header when it changes. It seems that we could do that in one run one of the migrator by setting the initial headers correctly?

test/integration/test_migrator.py
```diff
@@ -677,6 +681,41 @@ def key_gone(key):
...
+        conn_local.put_container(
+            'prop_metadata_test', headers={mykey: 'local'})
+        with self.migrator_running():
+            with self.assertRaises(WaitTimedOut):
+                wait_for_condition(3, partial(is_where, 'remote'))
+            self.assertTrue(is_where('local'))
+            conn_remote.post_container(
+                'prop_metadata_test', headers={mykey: 'remote'})
+            wait_for_condition(3, partial(is_where, 'remote'))
+            conn_local.put_container(
```
timuralp: I think this should be `post`? For PUT, we did behave correctly before, as `x-timestamp` would've been updated.

test/integration/test_migrator.py
```diff
@@ -677,6 +681,41 @@ def key_gone(key):
...
+            test_hdrs = conn_noshunt.head_container('prop_metadata_test')
+            if test_hdrs[mykey] == val:
+                return True
+            return False
+
+        conn_remote.put_container(
+            'prop_metadata_test', headers={mykey: 'remote'})
+        conn_local.put_container(
+            'prop_metadata_test', headers={mykey: 'local'})
+        with self.migrator_running():
```
timuralp: I'm wondering if this test could be simpler:
1. Create a source container and a destination container
2. POST on the source container, ensuring the last-modified date is ahead of the source container (probably have to sleep)
3. POST on the destination container, ensuring the last-modified date is ahead of the source-container's last-modified date
4. ensure no metadata is changed

Currently, we create containers with some headers, where the destination is >= source last-modified. Then make sure the headers are not copied. Then do a post and ensure the header is updated (this is already covered by a test previously). Then afterward we set the local header, wait to make sure it's not overwritten, then set the remote header *again* and wait for it to overwrite the header. This seems a bit redundant and every time we wait, takes some amount of time.

s3_sync/migrator.py
mdegerne: yep.

s3_sync/migrator.py
mdegerne: removing.

s3_sync/migrator.py
mdegerne: yep

test/integration/test_migrator.py
mdegerne: copy paste error, I'm certain. Thanks.

test/integration/test_migrator.py
mdegerne: Improved.

test/integration/test_migrator.py
mdegerne: modified.

test/integration/test_migrator.py
```diff
@@ -677,6 +674,38 @@ def key_gone(key):
...
+            test_hdrs = conn_noshunt.head_container('prop_metadata_test')
+            if test_hdrs[mykey] == val:
+                return True
+            return False
+
+        conn_remote.put_container(
+            'prop_metadata_test', headers={mykey: 'remote'})
+        conn_local.put_container(
+            'prop_metadata_test', headers={mykey: 'local'})
+        time.sleep(1)
```
timuralp: I think we talked about how 2 seconds was required before and it was racy otherwise? I'm curious what the resolution was. Definitely happy about 50% reduction in sleeping :)

PR #108, opened 2018-05-01
Bug/object compare timestamps

mdegerne:
Add test to verify the fix for containers also fixes metadata update propagation for objects.

### DIFF
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -706,6 +706,46 @@ class TestMigrator(TestCloudSyncBase):
                 wait_for_condition(3, partial(is_where, 'remote'))
             self.assertTrue(is_where('local'))
 
+    def test_object_metadata_copied_only_when_newer(self):
+        migration = self.swift_migration()
+        key = u'test_object-own'
+        content = 'test object'
+
+        where_header = 'x-object-meta-where'
+
+        conn_local = self.conn_for_acct(migration['account'])
+        conn_remote = self.conn_for_acct(migration['aws_account'])
+        conn_noshunt = self.conn_for_acct_noshunt(migration['account'])
+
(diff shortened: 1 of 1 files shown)


test/integration/test_migrator.py
```diff
@@ -716,6 +716,51 @@ def is_where(val):
...
+                raise
+            return False
+
+        conn_remote.put_object(
+            migration['aws_bucket'], key, StringIO.StringIO(content),
+            headers={where_header: 'remote'})
+        conn_local.put_object(
+            migration['container'], key, StringIO.StringIO(content),
+            headers={where_header: 'local'})
+        with self.migrator_running():
```
timuralp: I'm similarly concerned about this test as the container one. Walking through what it does again:
1. create two containers and put two objects, with differing metadata, with the destination object being after the source, so source.ts <= destination.ts.
2. wait 3 seconds to make sure the destination metadata is not overwritten
3. update source metadata and make sure the update propagates
4. update destination metadata and wait 3 seconds to make sure it is not overwritten
5. update source metadata and make sure the update propagates

It seems we test something twice and we are testing metadata propagation for which we already have a test. Similar to the container test, I'd suggest setting up the test as:
1. create two objects as is done currently
2. post on the source object (make sure it's after the destination last-modified date); post on the destination object
3. make sure the source metadata does not overwrite the destination metadata by running the migrator for 3 seconds.

By the way, it'd be great to add a TODO that we should explicitly call the migrator, as the 3 second timeout is hokey.

mdegerne: Fixed, I think.

mdegerne:
rebased

mdegerne:
Fixed test - should behave correctly now.

PR #109, opened 2018-05-02
Add container listings for cloud-connector

dbishop:
Also fix bug in splice_listings() where the limit param could be
exceeded.

Providers can now send extra headers with every request.
Cloud-connector uses this to bypass the cloud sync shunt when talking to
a remote (to c-c) Swift cluster via a provider.  This is implemented for
Swift providers too, though nothing uses that and it's not really
tested.

[#156747829]


s3_sync/cloud_connector/app.py
```diff
@@ -77,20 +81,81 @@ def __init__(self, app, local_to_me_profile, remote_to_me_profile, version,
...
+                # (fine) or some other error (also fine since that's more
+                # interesting than any 404).
+                headers = dict(filter_hop_by_hop_headers(
+                    remote_resp.headers.items()))
+                return swob.Response(app_iter=iter(remote_resp.body),
+                                     status=remote_resp.status,
+                                     headers=headers, request=req)
+            # This one does need to be an actual iterator and conform to the
+            # contract of yielding (None, None) when it's "done".
+            remote_iter = iter([(None, None)])
```
timuralp: It may be useful to log the error we received here and any additional information. Does it get logged in the provider?

s3_sync/cloud_connector/app.py
```diff
@@ -77,20 +81,81 @@ def __init__(self, app, local_to_me_profile, remote_to_me_profile, version,
...
+                    local_resp.headers.items()))
+                return swob.Response(app_iter=iter(local_resp.body),
+                                     status=local_resp.status,
+                                     headers=headers, request=req)
+            # This is ok because splice_listing() only iterates over the
+            # local_iter--it doesn't try to call next() or anything if it's
+            # empty.
+            local_iter = []
+
+        remote_resp, remote_iter = iter_listing(
```
timuralp: I wonder if at some point this should be a conditional. Paying the overhead on every container GET request may be expensive.

test/unit/test_sync_swift.py
```diff
@@ -363,6 +405,99 @@ def get_object(account, container, key, headers):
             mock.call('account', 'segment_container', 'slo-object/part2',
                       headers=swift_req_headers)])
 
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_upload_slo_extra_headers(self, mock_swift):
```
timuralp: I would not add this test, as we should mutate the `upload_slo` to use `_call_swiftclient`.

timuralp:
At the meta level, I'd really prefer this to be a smaller change. For example, all of the code refactoring bits could've been done in a separate commit, even on the same PR/branch. That would make it easy to skip the straightforward parts and focus on the more interesting parts. Maybe next time.

s3_sync/cloud_connector/app.py
dbishop: Ya, that'd be a neat feature to add later.

s3_sync/cloud_connector/app.py
dbishop: Good idea! Adding logging for any non-404 errors getting out of here.

s3_sync/sync_s3.py
```diff
@@ -367,15 +384,21 @@ def list_objects(self, marker, limit, prefix, delimiter=None,
...
-                             name=urllib.unquote(row['Key'])[key_offset:],
-                             last_modified=row['LastModified'].strftime(
-                             SWIFT_TIME_FMT),
-                             bytes=row['Size'],
-                             content_location=content_location,
-                             # S3 does not include content-type in listings
-                             content_type='application/octet-stream')
-                        for row in resp.get('Contents', [])]
+                # If list contents come from `swift3` running in a Swift
+                # cluster, the Etag header appears to include the double-quotes
```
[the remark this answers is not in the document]
dbishop: Fixed.  Speling is FUNdamental!

s3_sync/sync_swift.py
```diff
@@ -80,17 +84,20 @@ def upload_object(self, swift_key, policy, internal_client):
         if self._per_account and not self.verified_container:
             with self.client_pool.get_client() as swift_client:
                 try:
-                    swift_client.head_container(self.remote_container)
+                    swift_client.head_container(self.remote_container,
```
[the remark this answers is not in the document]
dbishop: Hashed out in person--I'd like to leave the extra headers consistently-applied and tested; when we reengineer it, the test can change too.

test/container/Dockerfile
```diff
@@ -29,4 +29,8 @@ COPY swiftclient.patch /tmp/
 RUN bash -c "cd /usr/local/src \
     && patch -p0 < /tmp/swiftclient.patch \
     && rm -f /tmp/swiftclient.patch"
+COPY swift3-unicode-fix.patch /tmp/
```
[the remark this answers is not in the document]
dbishop: /shrug

test/unit/test_sync_swift.py
dbishop: ditto to above.

timuralp:
Will look into the sporadically failing test later.

PR #110, opened 2018-05-02
Add ability to propagate account metadata

mdegerne:
Only for swift->swift migrations

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -35,7 +35,8 @@ from .provider_factory import create_provider
 from .utils import (convert_to_local_headers, convert_to_swift_headers,
                     get_container_headers, iter_listing, RemoteHTTPError,
                     SWIFT_TIME_FMT, diff_container_headers,
-                    get_sys_migrator_header, MigrationContainerStates)
+                    get_sys_migrator_header, MigrationContainerStates,
+                    diff_account_headers)
 from swift.common.http import HTTP_NOT_FOUND, HTTP_CONFLICT
 from swift.common import swob
 from swift.common.internal_client import UnexpectedResponse
@@ -312,7 +313,32 @@ class Migrator(object):
             self._maybe_delete_internal_container(local_container['name'])
             local_container = next(local_iterator)
 
+    def _process_account_metadata(self):
+        if self.config.get('protocol') != 'swift':
+            return
+        if not self.config.get('propagate_account_metadata'):
+            return
+
+        resp = self.provider.head_account()
+        if resp.status // 100 != 2:
+            raise UnexpectedResponse('Failed to read container headers for '
+                                     '"%s": %d' % (self.config['account'],
+                                                   resp.status_int), resp)
+        rem_headers = resp.headers
+        with self.ic_pool.item() as ic:
+            local_headers = self._head_internal_account(ic)
+            if rem_headers and local_headers:
+                header_changes = diff_account_headers(
+                    rem_headers, local_headers)
+                if header_changes:
+                    ic.set_account_metadata(self.config['account'],
+                                            dict(header_changes))
+                    self.logger.info(
+                        'Updated account metadata for %s: %s' %
+                        (self.config['account'], header_changes.keys()))
+
     def _next_pass(self):
+        self._process_account_metadata()
         worker_pool = eventlet.GreenPool(self.workers)
         for _ in xrange(self.workers):
             worker_pool.spawn_n(self._upload_worker)
@@ -391,6 +417,26 @@ class Migrator(object):
         for _ in range(self.workers):
             q.put(None)
         q.join()
+
+    def _head_internal_account(self, internal_client):
+        # This explicitly does not use get_account_metadata because it
+        # needs to be able to read the temp url key (swift_owner: True).
+        # It should be noted that this should be something internal client can
+        # do and there is a patch proposed to allow it.
+        req = swob.Request.blank(
+            internal_client.make_path(self.config['account']),
+            environ={
+                'REQUEST_METHOD': 'HEAD',
+                'swift_owner': True
+            })
+
+        resp = req.get_response(internal_client.app)
+        if resp.status_int // 100 != 2:
+            raise UnexpectedResponse('Failed to read container headers for '
+                                     '"%s": %d' % (self.config['account'],
+                                                   resp.status_int), resp)
+
+        return dict(resp.headers)
 
     def _update_container_headers(self, container, internal_client, headers):
         # This explicitly does not use update_container_metadata because it
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -81,6 +81,9 @@ class SyncSwift(BaseSync):
             'post_object', self.remote_container, swift_key,
             headers=headers)
 
+    def head_account(self):
+        return self._call_swiftclient('head_account', None, None)
+
     def put_object(self, swift_key, headers, body_iter, query_string=None):
         return self._call_swiftclient('put_object', self.container, swift_key,
                                       contents=body_iter, headers=headers,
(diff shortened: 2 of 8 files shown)


mdegerne:
Don't merge yet. Still needs to propagate account acls.

s3_sync/migrator.py
```diff
@@ -333,6 +357,26 @@ def _stop_workers(self, q):
...
+                'swift_owner': True
+            })
+
+        resp = req.get_response(internal_client.app)
+        if resp.status_int // 100 != 2:
+            raise UnexpectedResponse('Failed to read container headers for '
+                                     '"%s": %d' % (self.config['account'],
+                                                   resp.status_int), resp)
+
+        return dict([(k.lower(), v) for k, v in resp.headers.items()])
```
timuralp: Are we doing this because we don't want titlecased keys?

s3_sync/sync_swift.py
```diff
@@ -71,6 +71,17 @@ def _close_conn(conn):
         if conn.http_conn:
             conn.http_conn[1].request_session.close()
 
+    def head_account(self):
```
timuralp: Please add this to the common provider abstraction (`base_sync.py`).

s3_sync/sync_swift.py
```diff
@@ -71,6 +71,17 @@ def _close_conn(conn):
...
+        with self.client_pool.get_client() as swift_client:
+            try:
+                headers = swift_client.head_account()
```
timuralp: It would be better to call `_call_swiftclient`. Is there a reason we can't do that? I see plumbing there for handling `None` for container.

s3_sync/sync_swift.py
```diff
@@ -71,6 +71,17 @@ def _close_conn(conn):
...
+            except swiftclient.exception.ClientException as e:
```
timuralp: I'm not sure that the provider should hide the 404. I think the caller is responsible for dealing with the 404 in whatever way they think is appropriate.

s3_sync/migrator.py
```diff
@@ -285,7 +285,31 @@ def next_pass(self):
...
 
+    def _process_account_metadata(self):
+        if self.config.get('protocol') != 'swift':
+            return
+
+        rem_headers = self.provider.head_account()
+        with self.ic_pool.item() as ic:
+            local_headers = self._read_account_headers(ic)
+            if rem_headers and local_headers:
+                local_ts = _create_x_timestamp_from_hdrs(local_headers)
```
timuralp: I suspect this does not work at all for us. If I try to post to an account, there is no `Last-Modified` header and the `X-Timestamp` is not updated on a POST. Have you observed a different outcome in your experiments?

If my observation is correct, I suspect we can't take the same approach as with containers and objects. I would advocate for having the policy of source account headers always superseding the destination. We may need to talk to product about that.

s3_sync/migrator.py
```diff
@@ -333,6 +357,26 @@ def _stop_workers(self, q):
             q.put(None)
         q.join()
 
+    def _read_account_headers(self, internal_client):
```
timuralp: Maybe `_head_internal_account`?

s3_sync/utils.py
```diff
@@ -734,6 +734,27 @@ def diff_container_headers(remote_headers, local_headers):
                  if _propagated_hdr(k) or k in versioning_headers])
 
 
+def diff_account_headers(remote_headers, local_headers):
```
timuralp: Can we normalize all headers here to lowercase and avoid the mutation in `_read_account_headers`?

test/integration/test_migrator.py
```diff
@@ -587,6 +587,56 @@ def test_propagate_container_meta(self):
             'head_container', migration['container'])
         self.assertNotIn(u'x-container-meta-migrated-\u062a', new_hdrs)
 
+    def test_migrate_account_metadata(self):
+        migration = self.swift_migration()
+        # Note: we need to use noshunt, otherwise data is spliced by the shunt
```
timuralp: Is this true? My reading of the code says the account queries always pass through to the internal cluster. Did you observe differently?

test/integration/test_migrator.py
```diff
@@ -587,6 +587,56 @@ def test_propagate_container_meta(self):
...
+            return True
+
+        def validate_no_metadata():
+            hdrs = conn_local.head_account()
+            for hdr in expected_headers:
+                if hdrs.get(hdr) is not None:
+                    return False
+            return True
+
+        if not validate_no_metadata():
```
timuralp: This is confusing to me. Does the test assume some prior state? Is this because we don't clean up state?

s3_sync/utils.py
```diff
@@ -735,24 +738,31 @@ def diff_container_headers(remote_headers, local_headers):
 
 
 def diff_account_headers(remote_headers, local_headers):
+    # account acls need to be translated to sysmeta
+    def _fix_acl(key):
```
timuralp: Maybe not `fix` but `translate`?

s3_sync/migrator.py
mdegerne: Changed as suggested.

s3_sync/sync_swift.py
mdegerne: Done.

s3_sync/sync_swift.py
mdegerne: Agreed. Done.

s3_sync/migrator.py
mdegerne: Done.

s3_sync/utils.py
mdegerne: Done.

test/integration/test_migrator.py
mdegerne: removed spurious comment

test/integration/test_migrator.py
mdegerne: Explanation and TODO added.

s3_sync/utils.py
```diff
@@ -784,6 +787,36 @@ def diff_container_headers(remote_headers, local_headers):
...
+        if key == ACCOUNT_ACL_KEY:
+            return SYSMETA_ACCOUNT_ACL_KEY
+        return key
+
+    def _prop_hdr(hdr):
+        propagate_keys = [ACCOUNT_ACL_KEY, SYSMETA_ACCOUNT_ACL_KEY]
+        return hdr.startswith('x-account-meta') or hdr in propagate_keys
+
+    # remote_headers are unicode, local are unicode returns str
```
dbishop: What does "local are unicode returns str" mean?  If it means "local are utf8-encoded", can you change it to say that?

s3_sync/migrator.py
```diff
@@ -311,7 +312,32 @@ def _reconcile_containers(self):
             self._maybe_delete_internal_container(local_container['name'])
             local_container = next(local_iterator)
 
+    def _process_account_metadata(self):
+        if self.config.get('protocol') != 'swift':
+            return
+        if self.config.get('propagate_account_metadata') is not True:
```
dbishop: This is fine, I guess, but unless you specifically need different behavior for other falsy values (like `''`, '0', `[]`, etc.), I think this is usually written:
```
if not self.config.get('propagate_account_metadata'):
    ...
```

test/integration/test_migrator.py
```diff
@@ -587,6 +589,72 @@ def test_propagate_container_meta(self):
...
+
+        conn_remote.post_account({
+            'x-account-meta-test1': 'mytestval',
+            'x-account-meta-temp-url-key': 'mysecret',
+            u'x-account-meta-\u062atest1': u'mytestval\u062a',
+            ACCOUNT_ACL_KEY: acl,
+        })
+
+        self.assertTrue(validate_no_metadata())
+        with self.migrator_running():
```
dbishop: This would be a good candidate to rewrite to use the turn-the-migrator-crank-once stuff that @timuralp just added.  No sense adding new tests to code-base using the old pattern.

mdegerne:
Addressed individual comments. I also added a separate commit for getting rid of two of the "wait to see that nothing happens" tests.

s3_sync/sync_swift.py
timuralp: @mdegerne can you please add this method into the common abstraction?

mdegerne:
Done.

PR #111, opened 2018-05-04
Remove or prune containers deleted at the source during migrations

timuralp:
When migrating, if a container is removed from the origin, we may need to remove it from the destination. In some cases, the container should be preserved, but its contents pruned. Check the commit message for more details.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -259,6 +259,10 @@ class Migrator(object):
         try:
             self._reconcile_containers()
         except Exception:
+            # Any exception raised will terminate the migrator process. As the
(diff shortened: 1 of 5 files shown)


timuralp:
@mdegerne good question -- however, if we didn't create the container, we should never delete it. On the other hand, when we update container metadata, luckily it behaves differently from objects, where the sys meta headers will not be clobbered. Do you see a case where we may clear these headers and not delete/clear a container we should?

timuralp:
@mdegerne added a commit to document all of the bare exceptions in the core migrator code. I think you're right that it should be explained.

PR #112, opened 2018-05-07
Fixes a pagination bug when deleting objects

timuralp:
Fixes the pagination issue where we were previously removing objects we shouldn't have been. This would occur if the two blob stores returning listings with differing markers and we would assume that the fact that an object is missing from the listing means it should be removed, even though it might mean that the listing hasn't got as far yet.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -709,7 +709,8 @@ class Migrator(object):
                 scanned += 1
                 if remote:
                     marker = remote['name']
-        while local:
+
+        while local and (not marker or local['name'] < marker):
             # We may have objects left behind that need to be removed
             with self.ic_pool.item() as ic:
                 self._reconcile_deleted_objects(ic, container, local['name'])
--- a/test/container/swift-s3-sync.conf
+++ b/test/container/swift-s3-sync.conf
@@ -237,7 +237,7 @@
         }
     ],
     "migrator_settings": {
-        "items_chunk": 1000,
+        "items_chunk": 5,
         "log_file": "/var/log/swift-s3-migrator.log",
         "poll_interval": 1,
         "status_file": "/var/lib/swift-s3-sync/migrator.status",
--- a/test/integration/test_migrator.py
+++ b/test/integration/test_migrator.py
@@ -1090,3 +1090,44 @@ class TestMigrator(TestCloudSyncBase):
 
         with self.migrator_running():
             wait_for_condition(5, _check_deleted_migrated_objects)
+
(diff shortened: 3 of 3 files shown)


s3_sync/migrator.py
```diff
@@ -257,34 +258,54 @@ def next_pass(self):
...
-                    break
-                if index % self.nodes == self.node_id:
-                    # NOTE: we cannot remap container names when migrating all
-                    # containers
-                    self.config['aws_bucket'] = container['name']
-                    self.config['container'] = container['name']
-                    self.provider.aws_bucket = container['name']
-                    self._next_pass()
+            self._reconcile_containers()
         except Exception:
```
mdegerne: Bother's me to catch any exception here. Can we narrow down the type?

s3_sync/migrator.py
```diff
@@ -257,34 +258,54 @@ def next_pass(self):
...
+                # entire account
+                self.config['aws_bucket'] = remote_container
+                self.config['container'] = remote_container
+                self.provider.aws_bucket = remote_container
+                self._next_pass()
+            if local_container and local_container['name'] == remote_container:
+                local_container = next(local_iterator)
+
+        while local_container:
+            self._maybe_delete_internal_container(local_container['name'])
```
mdegerne: Is this not subject to the same limitation regarding pagination. Granted, it should be less common for > 10000 containers, but if we fix it one place, we should probably fix it here as well, no?

s3_sync/migrator.py
timuralp: This is on purpose -- happy to add a comment. The reason is that any exception here will prevent us from looking at a subsequent container. Rather, we log any exceptions and move on to the next container. There are a number of exceptions that could arise and I don't think it'd be useful to enumerate them here.

s3_sync/migrator.py
timuralp: This is somewhat different. We iterate through the full listing before we end up here. We don't use a marker for the prior state to restart and do not terminate listing after a fixed number of items (`work_chunk`).

timuralp:
@mdegerne I think we should finish reviewing/merge https://github.com/swiftstack/swift-s3-sync/pull/111, as I think the follow-on commit in this patch is fairly straightforward. Let me know if you'd rather prefer to review everything in this PR and I'll rebase it, as it's currently out of date.

s3_sync/migrator.py
mdegerne: Got it. Needed to read through the iter_listing code again to see what was going on. Thanks.

PR #114, opened 2018-05-10
Tooling to build a cloud-connector Docker container

dbishop:
Fixed issue with newer swift3 where cloud-connector's object GET 404
body could have the wrong message (bucket-not-found instead of
key-not-found, or whatever).

Unified how CI and developers run tests, including how the involved
containers are built and started.  The cloud-connector for tests is now
in its own container.

The use of docker-compose is mostly deprecated at this point, and when
switching to this code tree and beyond, you'll want to stop and "rm" the
docker-compose-related container, if you have one.

[#157237645]


s3_sync/cloud_connector/app.py
```diff
@@ -254,6 +294,8 @@ class CloudConnectorApplication(ProxyApplication):
     swift3 middleware) to run on cloud compute nodes and
     seamlessly provide R/W access to the "cloud sync" data namespace.
     """
+    modify_wsgi_pipeline = None
```
timuralp: Can you add a comment about why this is required?

README.md
```diff
@@ -92,70 +92,67 @@ This middleware should be in the pipeline before the DLO/SLO middleware.
 
 ### Trying it out
 
-If you have docker and docker-compose already you can easily get started in the root directory:
+Make sure you have docker installed and working.  Then run:
 
 ```
-docker-compose up -d
+./run_bash
```
timuralp: It might be nice to provider a small script that's equivalent to `docker-compose up -d`, as the prior behavior was that to run in the background. I don't think `run_bash` is the way to try it out.

README.md
```diff
@@ -92,70 +92,67 @@ This middleware should be in the pipeline before the DLO/SLO middleware.
...
+named `swift-s3-sync` and `cloud-connector`.  The `swift-s3-sync` container
+is based on the
+[bouncestorage/swift-aio](https://hub.docker.com/r/bouncestorage/swift-aio/) and
+includes a Swift all-in-one and a file-system-backed S3Proxy.  The `run_*` bash
+helper scripts map the
+source tree into the container, so that the cloud sync code operates on your
 current state.  Port 8080 is the Swift Proxy server, 10080 is the S3Proxy, and
 8081 is the Cloud Connector port.
 
+No ports are mapped to the host OS by default.
```
timuralp: Could we preserve the behavior of exposing swift and s3proxy? I've found it convenient to be able to talk to the services using tools outside of the container.

README.md
dbishop: ya, the ergonomics of this stuff sucks, but I guess I can head that direction.  I did not want to
* have a proliferation of too many different scripts to keep cognitive load down
* have useful things to run buried where they'd never get looked at

but that may not be worse thatn what's there now.

README.md
dbishop: I assumed the docker-compose override stuff was added b/c those ports hit existing stuff for some developer (@clayg maybe?).

I'm leaving it for future work to add that back in some way that works for developers who want it (you) and devs who _don't_ want it, or want it configurable (@clayg ?).  For me, I never added host-access to my workflow, so I don't have a horse in the race.

README.md
```diff
@@ -92,75 +92,88 @@ This middleware should be in the pipeline before the DLO/SLO middleware.
...
+includes a Swift all-in-one and a file-system-backed S3Proxy.  The `up` and
+`run_*` bash helper scripts map the
+source tree into the container, so that the cloud sync code operates on your
+current state.
+
+By default, the following services are pubished to host operating system ports:
+
+1. Swift: `8080`
+1. S3Proxy: `10080`
+1. Cloud Connector: `8081`
```
timuralp: Cool -- thanks for plumbing this through!

README.md
```diff
@@ -284,45 +236,98 @@ same Swift containers or S3 buckets.
...
-automatically overlay your customized ports:
+### Working with Docker directly
 
-`cp example.env .env`
+The best way to see how to build or run things in the docker containers
+is to examine the `run_*` helper shell scripts.
 
-Docker will automatically read the `.env` file and use your modified
-`docker-compose.custom.yml` to control the host port mapping.
+No ports are exposed on the host OS, so if you want to do that, you'll need
```
timuralp: I think you've plumbed this through now and we can talk to the exposed ports?

dbishop:
good catch on the vestigial doc thing; cleaned it up.

On Thu, May 17, 2018 at 11:51 AM, Timur Alperovich <notifications@github.com

PR #116, opened 2018-05-11
Add custom_prefix to ignored keys.

timuralp:
We should never use the custom_prefix field with migrations and
hard-code it to be an empty string. This is done to make sure migrations
out of S3 work correctly.


test/unit/test_migrator.py
```diff
@@ -489,6 +489,66 @@ def test_prune_all_buckets_migration(self):
...
+            json.dump(status_list, wf)
+        status = s3_sync.migrator.Status(self.status_file_path)
+        status.load_status_list()
+        migrator = s3_sync.migrator.Migrator(
+            config, status, 10, mock.Mock(max_size=1), None, 0, 1)
+        self.assertIn('custom_prefix', migrator.config)
+        self.assertEqual('', migrator.config['custom_prefix'])
+        status.save_migration(migrator.config, 'new-marker', 10, 100, False)
+        status.prune([config])
+        expected_status = [dict(config)]
```
mdegerne: Is "expected_status" just old code that needs to be cleaned up?

PR #120, opened 2018-05-18
cloud-connector: add support for DELETE'ing objects in S3 store

dbishop:
local-to-me objects can now be deleted; deletes are not propagated through
to the remote-to-me Swift store.

Also make provider delete_object() return a ProviderResponse

Also also, error ProviderResponses now have a `reraise()` method which will
reraise the low-level exception from the provider's library.  This is only
used to preserve some existing behavior, but it's potentially quite useful
for debugging, especially as providers are used more by cloud-connector,
etc.

### DIFF
--- a/s3_sync/base_sync.py
+++ b/s3_sync/base_sync.py
@@ -20,16 +20,30 @@ from swift.common.swob import RESPONSE_REASONS
 
 
 class ProviderResponse(object):
-    def __init__(self, success, status, headers, body):
+    def __init__(self, success, status, headers, body, exc_info=None):
         self.success = success
         self.status = status
         self.headers = headers
         self.body = body
+        # optional result from calling sys.exc_info(); used to implement
+        # reraise(), which allows a receiver of a ProviderResponse to reraise a
+        # prior client library exception.
+        self.exc_info = exc_info
 
     def to_wsgi(self):
         # WSGI expects an HTTP status + reason string
         wsgi_status = '%d %s' % (self.status, RESPONSE_REASONS[self.status][0])
         return wsgi_status, self.headers.items(), self.body
+
+    def __repr__(self):
+        return '<%s: %s, %r, %r, %r>' % (
+            self.__class__.__name__, self.success, self.status, self.headers,
+            (self.body or '')[:70])
+
+    def reraise(self):
+        if self.exc_info:
+            raise self.exc_info[0], self.exc_info[1], self.exc_info[2]
+        raise ValueError('reraise had no prior exception for %r' % self)
 
 
 class BaseSync(object):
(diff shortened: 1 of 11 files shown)


s3_sync/base_sync.py
```diff
@@ -20,17 +20,31 @@
...
 
     def to_wsgi(self):
         # WSGI expects an HTTP status + reason string
         wsgi_status = '%d %s' % (self.status, RESPONSE_REASONS[self.status][0])
         return wsgi_status, self.headers.items(), self.body
 
+    def __repr__(self):
+        return '<%s: %s, %r, %r, %r>' % (
+            self.__class__.__name__, self.success, self.status, self.headers,
+            (self.body or '')[:70])
```
timuralp: I think we should be careful about this, as in certain cases (on GET) it will mean returning partial response if `__repr__` gets called in any log line. I'd consider not including any part of the body here.

s3_sync/sync_s3.py
```diff
@@ -315,6 +315,12 @@ def _perform_op(s3_client):
             try:
                 resp = getattr(s3_client, op)(**args)
                 body = resp.get('Body', iter(['']))
+
+                # S3 API responses are inconsistent, so there will be various
+                # special-cases here.  This one about CopyObjectResult is for
+                # PUTs that copy objects (to update metadata or whatever).
+                # The API response for delete_object is also different (but
+                # only sometimes?)..
```
timuralp: Nit: double "."

Can you expand on when/how delete_object responses are different? Is this related to the NoBucketError vs 204 on an object that doesn't exist?

test/integration/test_cloud_connector.py
```diff
@@ -650,83 +650,127 @@ def test_container_get_filters_against_swift_dirs(self):
         ], list_resp['CommonPrefixes'])
         self.assertFalse(list_resp['IsTruncated'])
 
-    def test_obj_head_and_get(self):
-        # Not there yet...
+    def test_obj_head_and_get_and_delete(self):
```
timuralp: This is becoming a fairly long test to parse. Can we have a separate test for DELETE cases (objects in one, another, or both object stores)?

s3_sync/sync_s3.py
dbishop: No, I don't actually know why it's different sometimes.

test/integration/test_cloud_connector.py
dbishop: Weren't you just asking for more integration tests to be done in fewer functions for runtime efficiency??

I was trying to do what you wanted in the past...

s3_sync/base_sync.py
dbishop: ya, that was ill-advised; I'll do something better.

dbishop:
I did consider that, of course, but the setup to test DELETEs is basically
all the other crap that head/get test is doing, just without the asserts
related to head/get... so it'd almost literally be double the run-time
effort/cost.  The DELETE operations and assertions fit pretty well into the
existing actions taken by that existing testcase.  I agree it gets a bit
long-winded, but it flows reasonably.

PR #121, opened 2018-05-18
Use merge_namespaces explicitly in cloud_shunt

mdegerne:
Old way was to rely on propagate_delete setting. 

This will require existing config files to change to continue working as they have.


mdegerne:
You caught me, @dbishop. I really didn't think that would fly in under the radar, but I was wondering if anyone was looking. Working on it, still.

mdegerne:
I've added unit test for this.

s3_sync/shunt.py
```diff
@@ -111,8 +111,13 @@ def __init__(self, app, conf_file, conf):
 
         self.sync_profiles = {}
         for cont in conf.get('containers', []):
-            if cont.get('propagate_delete', True):
-                # object shouldn't exist in remote
+            mn = cont.get('merge_namespaces', None)
```
timuralp: Please use a better variable name than `mn`.

s3_sync/shunt.py
```diff
@@ -111,8 +111,13 @@ def __init__(self, app, conf_file, conf):
...
+            if mn is None:
+                if cont.get('propagate_delete', True):
```
timuralp: I don't think we need to worry about this. I think it's ok to require configuration to be updated with the new version of 1space.

s3_sync/shunt.py
```diff
@@ -111,8 +111,13 @@ def __init__(self, app, conf_file, conf):
...
+                    # implicitly not merging namespaces (old controller data)
```
timuralp: I would not explicitly mention the SwiftStack controller here, as 1space can be used outside of SwiftStack.

mdegerne:
Removed the implicit fall back to propagate delete setting at @timuralp's request.

dbishop:
Re: `We should require updating configuration files with this release.`, can we make sure that new code runs correctly on old config file?  I don't think python-ssnode's upgrade will drag it in, but nothing prevents the new code from coming in and seeing the old config (e.g. manual "yum upgrade" or equiv).  And even with a config push, I think swiftstack-swift-services gets upgraded before the new config gets sent out, guaranteeing that state, at least briefly.

timuralp:
Since we don't version config files, we'll end up with a situation where setting `propagate_delete: True` and omitting `merge_namespaces` will result in merging the namespaces (since we can't tell whether this is an old config file or an updated one that assumes namespaces are not merged unless explicitly stated). The failure mode is that namespaces would not be merged for old config files until they are updated. I don't think this is a show stopping issue and the resolution is clear (push a new config or update the files manually). Otherwise, we'll have confusing semantics when we're trying to disentangle them.

mdegerne:
Just rebased on latest master.

mdegerne:
OK, @dbishop , this error is weird. It's kind of like the service didn't set up correctly or something (note that it works fine locally). Any hints?

timuralp:
Looks like the fetching the cache timed out and subsequently importing the layers failed with a permission error:
```
adding /home/travis/docker to cache
creating directory /home/travis/docker
/home/travis/docker/40209c13f05e.tar.gz: Permission denied
```
I retriggered it to see if it's a spurious issue. We could harden the layer-import bit, but not sure how we get permission errors.

mdegerne:
Passing tests! Merge?

mdegerne:
@dbishop Given the other discussions, can I merge this?

timuralp:
I think we have resolved the issues that @dbishop brought up. I'm going to merge this patch.

PR #122, opened 2018-05-18
Allow for the migrator to be run once in tests.

timuralp:
When we write integration tests, it would be nice to have a tighter
level of control over the status and the migrator settings. The patch
adds an in-memory status class to be used with the migrator. It also
allows the migrator to be instantiated with specific configuration
settings, at which point it we can call next_pass() to get it to do a
pass over the configured migrations.

The patch mutates one test to take advantage of this approach, removing
the arbitrary sleep timer.


timuralp:
Reworked the tests to configure the number of manifests and segments based on the defined number of objects to consider at a time.

PR #123, opened 2018-05-21
Make shunt automatically reload configuration

mdegerne:
When the configuration file changes, the shunt will automatically reload
the configuration. This removes the requirement for restarting the proxy
server every time a change happens.


timuralp:
Looks good to me. I'd add a test before merging, though.

PR #130, opened 2018-05-24
Initialize the provider logger in the migrator.

mdegerne:


s3_sync/migrator.py
```diff
@@ -253,14 +253,14 @@ def __init__(self, config, status, work_chunk, workers, swift_pool, logger,
     def next_pass(self):
         if self.config['aws_bucket'] != '/*':
             self.provider = create_provider(
-                self.config, self.max_conns, False)
+                self.config, self.max_conns, False, logger=self.logger)
```
timuralp: I think it'd be better to follow the boto model here of configuring the handler and log level appropriately. The providers expect the log route to be named `s3-sync` (https://github.com/swiftstack/1space/blob/master/s3_sync/base_sync.py#L163), which means we can configure it similarly here https://github.com/swiftstack/1space/blob/master/s3_sync/daemon_utils.py#L48.

mdegerne:
I think this is the way to accomplish the goal - log size went from ~1.9 k lines to 24k lines.

timuralp:
@mdegerne can you elaborate on why opting into the existing logging framework way of doing this is not preferred? I don't doubt that it works, but question whether we want to be passing logger objects around, especially when we already do the same kind of setup in `setup_logger()` for boto3 and botocore.

mdegerne:
@timuralp Please check the code change I submitted before my comment. I implented, I think, what you are asking for.

timuralp:
@mdegerne I'm sorry -- I made a mistake! did not notice that code changed and thought you were responding to the first comment. I'll have to be sure to check the code in the future, as it's not obvious sometimes whether changes were made.

Do you think we could do this setup in `setup_logger` itself? Would there be any negative consequences to that? Also, can you please add the call `setLevel`, so that `s3-sync` logs at the same severity.

mdegerne:
I was primarily being lazy by not looking what else used setup_logger (and it looks like there is no reason not to do it all there). Oops on the setLevel. 

Fixed.

timuralp:
@mdegerne awesome -- thanks! Sorry for the back and forth on this. This patch looks great!

PR #131, opened 2018-05-30
Do not restart a stopped container.

timuralp:
Restarting the cloud connector when it's stopped makes it less obvious
to remove the container.

### DIFF
--- a/scripts/ensure_cloud_connector_container_started
+++ b/scripts/ensure_cloud_connector_container_started
@@ -24,7 +24,7 @@ docker container rm cloud-connector 2>/dev/null ||:
 
 docker run -d -v `pwd`:/swift-s3-sync \
     --network swift-s3-sync-net --network-alias cloud-connector \
-    --restart on-failure --name cloud-connector \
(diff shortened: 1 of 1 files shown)


dbishop:
I don't get it, when does it need to be stopped?

timuralp:
@dbishop when it's running on my laptop. Once the containers are running, rebuilding this one and setting it up again has to be done with `docker rm -f`. Is there a reason we want the container to be restarted when explicitly stopped by `docker stop`?

dbishop:
@timuralp if you want to rebuild the container, you run `ensure_cloud_connector_container_started` which does a stop and rm.  IME, that always worked for me...

timuralp:
I don't see a negative consequence to not restarting on explicit `docker stop` events, although I now can't reproduce the original problem I bumped into. Will reopen if I bump into it.

When I observed the restart behavior, `ensure_cloud_connector_container_started` is racy, because if the connector is restarted before `docker rm`, it will fail.

PR #134, opened 2018-06-07
Better Distribution of Migration Work

mdegerne:
This distributes the work using the container ring (assumption being that the container nodes all run migrator).


s3_sync/migrator.py
```diff
@@ -349,28 +354,38 @@ def _process_account_metadata(self):
                         'Updated account metadata for %s: %s' %
                         (self.config['account'], header_changes.keys()))
 
-    def _next_pass(self):
+    def _next_pass(self, is_verify):
```
timuralp: I wonder if we should call `_next_pass()` twice, as opposed to handling the verification at the end of the one call to `_next_pass()` (since we won't have to shut down and recreate the worker pools that way and the verification state tracking and calls would be slightly less awkward).

s3_sync/migrator.py
```diff
@@ -349,28 +354,38 @@ def _process_account_metadata(self):
...
+                # We must catch any errors to make sure we stop our workers.
+                # This might be better with a context manager.
+                scanned = 0
+                self.logger.error('Failed to migrate "%s"' %
+                                  self.config['aws_bucket'])
+                self.logger.error(''.join(traceback.format_exc()))
+
+        else:
+            scanned = copied = 0
         self.object_queue.join()
```
timuralp: Follow on to the suggestion about having a single call to the verification mechanism is that we could have a loop that drains the `verify_queue` and puts the work items into the `object_queue` after calling `join` on the `object_queue`, draining the `container_queue`, and handling the manifests set (repeating these actions, maybe in a helper method).

s3_sync/migrator.py
```diff
@@ -789,7 +811,13 @@ def _find_missing_objects(
                 if self._old_enough(remote):
                     work = MigrateObjectWork(aws_bucket, container,
                                              remote['name'])
-                    self.object_queue.put(work)
+                    if list_all:
```
timuralp: Are we doing this to take care of the versioned container? I think versioning now is tricky. There are likely N migrator processes responsible for the versions container. In addition, since versioning is per-container, there will be N migrator processes trying to do it with the full scan option. Maybe we have to change our guarantee on versioned containers and not try to process them ahead of time?

test/unit/test_migrator.py
```diff
@@ -761,7 +776,7 @@ def test_migrate_objects(self, create_provider_mock):
 
         tests = [{
             'objects': {
-                'foo': {
+                'foo1': {
```
timuralp: I think these changes were done to disambiguate the tests. Maybe we can use `enumerate()` and include the test case index on failure?

timuralp:
@mdegerne a meta comment: could you expand on the commit messages a little bit? While I know the context right now, I know that when I look back on this in a few months, the details of why we implemented this scheme the way we did may be hazy. It'd be very helpful to have a short explanation of what about and why we're changing the way we distribute migration work across nodes.

s3_sync/migrator.py
```diff
@@ -1040,21 +1055,19 @@ def main():
     swift_dir = conf.get('swift_dir', '/etc/swift')
     internal_pool = create_ic_pool(conf, swift_dir, workers)
 
-    if 'process' not in migrator_conf or 'processes' not in migrator_conf:
-        print 'Missing "process" or "processes" settings in the config file'
-        exit(-1)
+    container_ring = Ring(swift_dir, ring_name='container')
```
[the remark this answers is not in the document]
mdegerne: In SwiftStack, it is always container, but I can easily imagine a scenario where someone chooses to run the migrators on a different set of nodes and may want to use a different ring, so I'll add configuration option for that.

s3_sync/migrator.py
timuralp: I think avoiding stopping/spinning up green threads should be our goal in this patch. Every time we do that, we have to terminate and re-create the boto connections. I don't think the overhead of copying object names from the deferred queue into the `object_queue()` would be that significant. We should probably factor out the loops that deal with `container_queue` and `_manifests` into a separate method and call it after we put all the verified objects into the queue. I think that would look a bit cleaner, as well as avoid resetting all the connection pools.

s3_sync/migrator.py
```diff
@@ -63,6 +66,27 @@
 S3_MPU_RE = re.compile('[0-9a-z]+-(\d+)$')
 
 
+def is_local_container(myips, ring, primary_only, acct, cont, obj=None):
+    if obj:
```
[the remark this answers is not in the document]
mdegerne: is_primary is a partial for this and we call is_primary with an object.

test/unit/test_migrator.py
mdegerne: fixed

test/unit/test_migrator.py
```diff
@@ -1348,7 +1383,8 @@ def fake_get_metadata(account, container):
                 self.swift_client.container_exists.side_effect = (False, True)
                 headers = {}
 
-            self.swift_client.iter_objects.return_value = iter([])
+#            self.migrator._iterate_internal_listing = \
```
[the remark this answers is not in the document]
mdegerne: fixed

s3_sync/migrator.py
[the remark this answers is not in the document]
mdegerne: fixed - finally

PR #138, opened 2018-06-14
Use x-timestamp when deleting migrated objects

mdegerne:
This prevents a very narrow window for a new copy of the object to be put
and then deleted incorrectly.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -547,10 +547,30 @@ class Migrator(object):
                     return
                 raise
             if get_sys_migrator_header('object') in hdrs:
-                self.logger.info(
-                    'Detected removed object %s. Removing from %s/%s' % (
-                        key, self.config['account'], container))
-                ic.delete_object(self.config['account'], container, key)
+                headers = {}
+                xts = hdrs.get('x-backend-durable-timestamp') or \
+                    hdrs.get('x-backend-timestamp') or hdrs.get('x-timestamp')
+                if xts:
+                    ts = Timestamp(xts)
+                    headers['x-timestamp'] = Timestamp(
+                        ts.timestamp, ts.offset + 1).internal
+                else:
+                    xts = _create_x_timestamp_from_hdrs(hdrs)
+                    if xts:
+                        headers['x-timestamp'] = Timestamp(xts, 1)
+                try:
+                    ic.delete_object(self.config['account'], container, key,
+                                     headers)
+                    self.logger.info(
+                        'Detected removed object %s. Removing from %s/%s' % (
+                            key, self.config['account'], container))
+                except UnexpectedResponse as e:
+                    if e.resp.status_int == HTTP_CONFLICT:
+                        self.logger.info(
+                            'Conflict removing object %s from %s/%s' % (
+                                key, self.config['account'], container))
+                        return
+                    raise
 
     def _maybe_delete_internal_container(self, container):
         '''Delete a specified internal container.
--- a/test/unit/test_migrator.py
+++ b/test/unit/test_migrator.py
@@ -1773,6 +1773,56 @@ class TestMigrator(unittest.TestCase):
             {internal_header: '1500000000.00000',
              'x-timestamp': '1500000000.00000'})
 
+    def test_reconcile_deleted_timestamps(self):
+        internal_header = s3_sync.utils.get_sys_migrator_header('object')
+        tests = [
+            ({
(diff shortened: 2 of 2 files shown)


s3_sync/migrator.py
```diff
@@ -546,10 +546,22 @@ def _reconcile_deleted_objects(self, container, key):
                     return
                 raise
             if get_sys_migrator_header('object') in hdrs:
+                headers = {}
+                xts = hdrs.get('x-timestamp')
+                if xts:
+                    ts = Timestamp(xts)
+                    headers['x-timestamp'] = repr(Timestamp(
+                        ts.timestamp, ts.offset + 1))
                 self.logger.info(
```
timuralp: I would move this log line in the `try` clause, after `delete_object` succeeds.

s3_sync/migrator.py
```diff
@@ -546,10 +546,22 @@ def _reconcile_deleted_objects(self, container, key):
                     return
                 raise
             if get_sys_migrator_header('object') in hdrs:
+                headers = {}
+                xts = hdrs.get('x-timestamp')
+                if xts:
+                    ts = Timestamp(xts)
+                    headers['x-timestamp'] = repr(Timestamp(
+                        ts.timestamp, ts.offset + 1))
```
timuralp: We should call `.internal` here to avoid being bit by any changes to how `__repr__` is implemented.

Can you check if the current offset appears in the `x-timestamp` header? If not, is there any we can retrieve it? I'm worried that our deletes may fail if we don't get the current offset correctly.

s3_sync/migrator.py
```diff
@@ -546,10 +546,22 @@ def _reconcile_deleted_objects(self, container, key):
...
                 self.logger.info(
                     'Detected removed object %s. Removing from %s/%s' % (
                         key, self.config['account'], container))
-                ic.delete_object(self.config['account'], container, key)
+                try:
+                    ic.delete_object(self.config['account'], container, key,
+                                     headers)
+                except UnexpectedResponse as e:
+                    if e.resp.status_int == HTTP_CONFLICT:
```
timuralp: This should be uncommon -- let's log an INFO-level message about the conflict.

s3_sync/migrator.py
mdegerne: fixing.

s3_sync/migrator.py
mdegerne: If an offset exists for the object, the `x-backend-durable-timestamp` has the full timestamp value with offset, so does the `x-backend-data-timestamp`. I'm going to experiment a bit to see how these change with updates.

s3_sync/migrator.py
mdegerne: fixing

s3_sync/migrator.py
```diff
@@ -545,10 +545,33 @@ def _reconcile_deleted_objects(self, container, key):
                     return
                 raise
             if get_sys_migrator_header('object') in hdrs:
-                self.logger.info(
-                    'Detected removed object %s. Removing from %s/%s' % (
-                        key, self.config['account'], container))
-                ic.delete_object(self.config['account'], container, key)
+                headers = {}
+                xts = hdrs.get(
```
timuralp: Nit/style comment; what do you think about using `or` here instead?
```
xts = hdrs.get('x-backend-durable-timestamp') or hdrs.get('x-backend-timestamp') or hdrs.get('x-timestamp')
```

The small, meaningless here benefit, is that we avoid evaluating every `get()` call.

test/unit/test_migrator.py
```diff
@@ -1772,6 +1772,40 @@ def test_reconcile_deleted_object(self, create_provider_mock):
...
+            self.migrator._reconcile_deleted_objects('foo', 'bar')
+
+            hdrs = {}
+            if want:
+                hdrs = {'x-timestamp': want}
+            self.assertEqual(
+                self.swift_client.delete_object.call_args,
+                mock.call(self.migrator.config['account'], 'foo', 'bar', hdrs))
+            if use:
+                del(obj_headers[use])
```
timuralp: I get what we're trying to here, but I think mutating the object headers is a little difficult to read. Can you change the test to set the object headers explicitly to the values we want to test?

s3_sync/migrator.py
mdegerne: I like this form better. I was trying to find a good way to write it.

PR #139, opened 2018-06-25
Migrator: Remove an extra GET request for S/DLOs.

timuralp:
When migrating large objects, we can avoid an extra GET request.
Previously, we would do a generic GET, check if an object is a large
object, then submit another GET request. The patch changes the behavior
to always submit "multipart-manifest=get" with every object GET request.
For regular objects, Swift ignores the query parameter. For large
objects, we get back the manifest.


timuralp:
@tipabu I believe this patch implements your suggestion to remove an extra (possibly pretty expensive) GET.

s3_sync/migrator.py
```diff
@@ -841,43 +845,41 @@ def _migrate_object(self, aws_bucket, container, key):
                 aws_bucket))
             return
 
-        resp = self.provider.get_object(key, **args)
-        if resp.status != 200:
-            resp.body.close()
-            raise MigrationError('Failed to GET "%s/%s": %s' % (
-                aws_bucket, key, resp.body))
-        put_headers = convert_to_local_headers(
-            resp.headers.items(), remove_timestamp=False)
```
tipabu: It seems weird to push this down into `_migrate_dlo`/`_migrate_slo`/the `else` block...

s3_sync/migrator.py
```diff
@@ -841,43 +845,41 @@ def _migrate_object(self, aws_bucket, container, key):
...
         if dlo_container != container or not key.startswith(prefix):
             # The DLO prefix can include the manifest object, which doesn't
             # have to be 0-sized. We have to be careful not to end up recursing
             # infinitely in that case.
             self._manifests.add((aws_bucket, container, key))
+        resp.body.close()
+
+    def _migrate_slo(self, aws_bucket, slo_container, key, resp):
+        manifest_blob = FileLikeIter(resp.body).read()
+        manifest = json.loads(manifest_blob)
```
tipabu: We can't just use `manifest = json.loads(resp.body)`? Why add the `FileLikeIter`?

s3_sync/migrator.py
```diff
@@ -841,43 +845,41 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        put_headers = convert_to_local_headers(
+            resp.headers.items(), remove_timestamp=False)
+        resp.body.close()
 
-    def _migrate_slo(self, aws_bucket, slo_container, key, headers):
-        manifest = self.provider.get_manifest(key, aws_bucket)
```
tipabu: Is `get_manifest` still used anywhere? How/for what?

s3_sync/migrator.py
```diff
@@ -841,43 +845,41 @@ def _migrate_object(self, aws_bucket, container, key):
...
             # infinitely in that case.
             self._manifests.add((aws_bucket, container, key))
+        resp.body.close()
+
+    def _migrate_slo(self, aws_bucket, slo_container, key, resp):
+        manifest_blob = FileLikeIter(resp.body).read()
+        manifest = json.loads(manifest_blob)
+        put_headers = convert_to_local_headers(
+            resp.headers.items(), remove_timestamp=False)
+        resp.body.close()
```
tipabu: I think this should already happen (assuming `resp` is a `swob.Response`) -- see https://github.com/openstack/swift/blob/2.18.0/swift/common/swob.py#L309-L310

Or are we worried about older versions of Swift? Looks like I only fixed that back in 2.15.0 -- https://github.com/openstack/swift/commit/0b60b0bd7ebfa84007515cc76cc5077e83cf2767

s3_sync/migrator.py
```diff
@@ -821,14 +820,19 @@ def _find_missing_objects(
         return marker, scanned, copied
 
     def _migrate_object(self, aws_bucket, container, key):
-        args = {'bucket': aws_bucket}
+        args = {'bucket': aws_bucket,
+                'query_string': 'multipart-manifest=get'}
```
tipabu: Should this also have `&format=raw` so it'll be in a format that we can re-upload? Do we care about pre-2.7.0 swift? https://github.com/openstack/swift/commit/e1f12960884f9ea52416891a8246a53f9673348b

No, I guess the translation is handled in `_upload_slo`...

timuralp: We do not at the moment.

s3_sync/migrator.py
timuralp: I found it weird not to do that, since we're passing the response into those methods and they can decide what to do with the headers (even if it's calling the same thing).

s3_sync/migrator.py
timuralp: No, because ClosingResourceIterable does not implement `read()`.

s3_sync/migrator.py
timuralp: It is a `ProviderResponse`, which wraps the SwiftClient `_RetryBody`.

s3_sync/migrator.py
timuralp: Yes, in the cloud sync code.

PR #146, opened 2018-07-11
Security Bug AWS password in debug log

mdegerne:
This replaces secret with "(redacted)"


mdegerne:
Tested locally - may still want tests, just not sure how best to do them.

notmyname:
alternatively, we could do something similar as upstream swift (and even use the same method)

notmyname:
ah, it was https://review.openstack.org/#/c/548948/7/swift/common/middleware/proxy_logging.py that I was thinking more of. Unfortunately not landed yet.

timuralp:
@mdegerne looks good. I think we can write a unit test fairly easily that verifies that the secret is replaced using a logger mock?

mdegerne:
I'll squash, just wanted to make it easy to read.

PR #147, opened 2018-07-12
Remove the log line when shunt is missing config.

timuralp:
The shunt will log on every request if its configuration file is
missing. We should remove the warning, as it can be disconcerting to the
operators.


mdegerne:
Is there any way we can log this once, when it runs into a problem? I really want to have some information in the logs saying that the file is unreadable or corrupted.

timuralp:
@mdegerne I don't think running this error from the middleware is as helpful. The same configuration is used by the 1space daemons and we would get errors through that. There is one set of errors where the middleware might not be able to read this file, but the daemons can. It'd be great if we come up with a way to check the configuration file in that case, however we'd need to investigate how to make sure it doesn't produce a log line on every request. I'd defer that investigation, though, for now.

mdegerne:
@timuralp Fair enough.

PR #148, opened 2018-07-12
Migrator: report the size of copied objects.

timuralp:
When the migrator copies objects, aside from reporting the total
numbers, we should try to keep track of their size. This allows
reasoning about the data throughput when trying to understand the
network/disk utilization by the migrator. The change adds two entries to
the status files: bytes_count and last_bytes_count. They are updated on
each pass.

To be able to keep track of the size of the objects uploaded and a more
accurate count of total work, the counter tracking has been moved into
the coroutines.


timuralp:
@mdegerne yep, I found the documentation lacking as well. The closest comparison is the `_threading_local` documentation: https://github.com/python/cpython/blob/master/Lib/_threading_local.py

I'd like to keep this open for a little bit in case @clayg wants to take a look.

PR #149, opened 2018-07-12
Bug fix for unicode handling of s3_key

mdegerne:
The key was not being encoded correctly if there was a custom prefix.


s3_sync/sync_s3.py
```diff
@@ -698,7 +698,11 @@ def get_prefix(self):
     def get_s3_name(self, key):
         prefix = self.get_prefix()
         if self.use_custom_prefix:
-            return '/'.join(filter(None, (prefix, key)))
+            if isinstance(key, unicode):
```
timuralp: Maybe this is simpler?
```
if not isinstance(key, unicode):
    key = key.decode('utf-8')
```

test/container/swift-s3-sync.conf
```diff
@@ -20,6 +20,7 @@
             "aws_secret": "s3-sync-test",
             "container": "\u062acl-conn-sync",
             "copy_after": 0,
+            "prefix": "pre-",
```
timuralp: Is `prefix` a valid field? I only see references to `settings.get('custom_prefix')`

s3_sync/sync_s3.py
mdegerne: yes, I was thinking for some reason that strings were pass by reference and I would be modifying. - Not true.

test/container/swift-s3-sync.conf
mdegerne: Aargh - fooled by old/invalid test data. Fixing.

PR #150, opened 2018-07-19
Ignore ProxyFS ETags during migrations.

timuralp:
Normally, we include the ETag to validate that our intended uploaded
data has been transferred correctly. Since ProxyFS does not allow us to
get the content hash, we have to ignore the ETag.


test/unit/test_utils.py
```diff
@@ -82,13 +82,23 @@ def test_convert_to_local_headers(self):
...
+                 'etag': 'deadbeef'}.items(),
                 remove_timestamp=False))
 
+        self.assertEqual(
+            {'content-type': 'application/test'},
+            utils.convert_to_local_headers(
+                {'x-timestamp': 12345,
+                 'etag': '"pfs-etag-some-value"',
+                 'content-type': 'application/test'}.items()))
+
```
mdegerne: I'm confused as to why the x-timestamp disappears in this new case.

timuralp: That's because the function is written as:
```
def convert_to_local_headers(headers, remove_timestamp=True)
```
and I was lazy and didn't pass `remove_timestamp=False`, so it got stripped out. Agreed that it's confusing. I can remove `x-timestamp` from the header set to be converted.

timuralp:
@mdegerne agreed! I struggled with that, as well. Here is my thought process: some ETags, like SLO and ones from S3, are quoted, but other ETags may not be quoted (like the ones Swift uses, even if they are violating the RFC -- https://tools.ietf.org/html/rfc7232#page-9). I think it's better to not include the " in the prefix, as the quotation should be a part of just being a valid ETag. At least that's how I think of it.

PR #151, opened 2018-07-21
Migrator: handle corrupted status files.

timuralp:
Status files may be corrupted if the migrator is killed in the process
of writing then. The change makes two changes:
1. status files are now written to a temporary file and then renamed
2. if the status file is corrupted, the migrator moves it with a
   corrupted suffix and restarts

While the first change may be sufficient to prevent the situations of
the processing being killed in the middle of a write, we should still be
robust to something mutating the file as it is stored on disk.


timuralp:
@clayg I thought you'd enjoy removing some of the mocks in lieu of using the temporary directories (it's somewhat difficult to mock at the right level when dealing with these temporary files). Also, found a small bug where `os.mkdir` does not accept keyword arguments, despite what the documentation says. Turned it up by actually calling `os.mkdir` -- victory for real tests over mocks in this case.

PR #152, opened 2018-07-26
Verify command without password on command line

mdegerne:
Modify the verification code to not pass the password on the command line
where it can be snooped via process listing.

This should be merged with a matching change to the controller. <Update link>


dbishop:
Does it still work ok when run by hand?  (does it need to?)

I'm thinking about 1) password getting echoed to screen and 2) needing to terminate the stdin stream for the CLI invocation to complete?

mdegerne:
It does not need to run by hand, and if anyone does, it should be `echo
<passwd> | ...`, which is silly, so I would recommend that if you want to
run it by hand, you put the password in a file and pass that file name.

timuralp:
Running this by hand may actually be useful, especially when debugging. If we make this work with `getpass` (https://docs.python.org/2/library/getpass.html) this could be better.

dbishop:
Does getpass work w/stdin?  If not, you'd need to look for something like
presence of a tty or some other indicator of interactive usage before doign
getpass vs. a read of stdin.

My question boiled down to, "Is that extra logic worth it, noting that we
need to test both cases, keep them both working, etc."

test/integration/test_verify.py
```diff
@@ -31,33 +33,47 @@ def setUp(self):
                 self.s3_bucket = container['aws_bucket']
             if self.swift_container and self.s3_bucket:
                 break
+        fd, self.tmpfile = tempfile.mkstemp()
+        os.close(fd)
+
+    def tearDown(self):
+        os.unlink(self.tmpfile)
```
tipabu: nit: Might want to be defensive and

- set `self.tmpfile = None` at the start of `setUp` and
- do an `if self.tmpfile:` check before unlinking

in case there's an error during `setUp`

test/unit/test_verify.py
```diff
@@ -24,13 +25,19 @@
 
 @mock.patch('s3_sync.verify.validate_bucket', return_value=object())
 class TestMainTrackProvider(unittest.TestCase):
+    def setUp(self):
+        self.stdin = sys.stdin
+        sys.stdin = StringIO('secret key')
+
+    def tearDown(self):
+        sys.stdin = self.stdin
```
tipabu: I wonder if this (and the similar code for `TestMainTrackClientCalls`) would look better with a `mock.patch`...

mdegerne: I agree, but I tried several unsuccessful variants of that before settling on this as a working solution. If you have a working example of mocking stdin to use, let me know. I would be happy to replace this with a better solution even if it means I overlooked something obvious.

timuralp:
Do we possibly want to change this to use `getpass()` and merge in?

mdegerne:
switched to getpass() instead of stdin.

PR #154, opened 2018-08-03
Handle self-referencing DLOs correctly.

timuralp:
If a DLO references itself, we would end up in an infinite recursive
loop, as we will keep trying to process its segments.

The patch checks whether a DLO manifest has already been seen and only
then will add the the required container/prefix to the container_queue.


timuralp:
@mdegerne I'd like to package this up for eBay tomorrow morning. If you get a chance to look at this before then, that'd be great!

s3_sync/migrator.py
```diff
@@ -906,7 +906,8 @@ def _migrate_object(self, aws_bucket, container, key):
                 aws_bucket))
             return
 
-        if 'x-object-manifest' in resp.headers:
+        if 'x-object-manifest' in resp.headers and\
+                (aws_bucket, container, key) not in self._manifests:
             self.logger.warning(
```
mdegerne: Does this not end up with the manifest file falling through to the regular uploadObject? Or is that what you want?

PR #157, opened 2018-08-15
Fix Keystone auth when using remote_account

dbishop:
Previously, when a swift protocol mapping using "remote_account", the
storage URL would have the keystone server's hostname and port, which
won't work.

The storage URL returned by keystone is the URL that needs to be munged
to have the overridden remote_account value.

Also beefed up Keystone integration tests a bit.

### DIFF
--- a/containers/swift-s3-sync/tempauth-keystone-groups.patch
+++ b/containers/swift-s3-sync/tempauth-keystone-groups.patch
@@ -1,0 +1,15 @@
+diff -ru swift.orig/swift/common/middleware/tempauth.py swift/swift/common/middleware/tempauth.py
+--- swift.orig/swift/common/middleware/tempauth.py	2018-08-13 23:33:28.000000000 +0000
++++ swift/swift/common/middleware/tempauth.py	2018-08-15 18:50:05.159165000 +0000
+@@ -547,7 +547,10 @@
+                 headers = [('Content-Type', 'text/plain; charset=UTF-8')]
+                 return HTTPBadRequest(request=req, headers=headers, body=msg)
+ 
+-        user_groups = (req.remote_user or '').split(',')
++        if isinstance(req.remote_user, (list, tuple)):
++            user_groups = [g.encode('utf8') for g in req.remote_user]
++        else:
++            user_groups = (req.remote_user or '').split(',')
+         account_user = user_groups[1] if len(user_groups) > 1 else None
+ 
+         if '.reseller_admin' in user_groups and \
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -45,27 +45,18 @@ class SyncSwift(BaseSync):
 
     def _get_client_factory(self):
         # TODO: support LDAP auth
-        # TODO: support v2 auth
         username = self.settings['aws_identity']
         key = self.settings['aws_secret']
         # Endpoint must be defined for the Swift clusters and should be the
         # auth URL
         endpoint = self.settings['aws_endpoint']
-        os_options = {}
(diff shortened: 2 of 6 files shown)


containers/swift-s3-sync/tempauth-keystone-groups.patch
```diff
@@ -0,0 +1,15 @@
...
++++ swift/swift/common/middleware/tempauth.py	2018-08-15 18:50:05.159165000 +0000
+@@ -547,7 +547,10 @@
+                 headers = [('Content-Type', 'text/plain; charset=UTF-8')]
+                 return HTTPBadRequest(request=req, headers=headers, body=msg)
+ 
+-        user_groups = (req.remote_user or '').split(',')
++        if isinstance(req.remote_user, (list, tuple)):
++            user_groups = [g.encode('utf8') for g in req.remote_user]
++        else:
++            user_groups = (req.remote_user or '').split(',')
```
mdegerne: req.remote_user needs to be encoded in one case and not the other???

dbishop: AFAIK, yes.

In the first branch, the values come from some Keystone-related auth middleware and are unicode instances in a tuple.

In the 2nd branch, it's whatever string was stuck in REMOTE_USER by tempauth itself, which I sort of assumed was already a UTF8-encoded values comma separated.  Is that not the case?

dbishop: ya, if the groups came out of memcached, they were Unicode and then encoded (~line 442)

```
                groups = groups.encode('utf8')
```
otherwise they come from 
```
        groups = [account, account_user]
        groups.extend(self.users[account_user]['groups'])
        if '.admin' in groups:
            groups.remove('.admin')
            for prefix in self._non_empty_reseller_prefixes():
                groups.append('%s%s' % (prefix, account))
            if account_id not in groups:
                groups.append(account_id)
        groups = ','.join(groups)
        return groups
```
which I guess is not encoding?  but not sure if they were unicode or not

s3_sync/sync_swift.py
```diff
@@ -90,7 +81,22 @@ def _get_client_factory(self):
             connection_kwargs['auth_version'] = '3'
 
         def swift_client_factory():
-            return swiftclient.client.Connection(**connection_kwargs)
+            _conn = swiftclient.client.Connection(**connection_kwargs)
+            if self.settings.get('remote_account'):
+                # We need to rewrite the account portion of the connection's
+                # storage URL (NOTE: this wouldn't't survive a
```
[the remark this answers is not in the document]
dbishop: YOU WOULDN'T'T SPELLCHECK A CAR WOULD YOU?!

test/integration/__init__.py
```diff
@@ -384,6 +404,16 @@ def setUpClass(klass):
...
 
+        # A test is going to need this Keystone user (tenant? whatever) able to
+        # read/write this tempauth account.
+        keystone_uuid = url_user_key_to_acct[(
+            "http://1space-keystone:5000/v2.0",
+            "tester", "testing", "test")].replace('KEY_', '')
+        acl_dict = {'read-write': [keystone_uuid]}
+        acl = format_acl(version=2, acl_dict=acl_dict)
+        conn = klass.conn_for_acct(u"AUTH_\u062aacct2")
+        resp_headers, body = conn.post_account({ACCOUNT_ACL_KEY: acl})
```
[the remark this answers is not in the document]
dbishop: haha, ya i had debugging shit on there when the acl wasn't sticking (i was using the noshunt proxy which is also no-tempauth, so no one persisted the acl).  #willfix

PR #159, opened 2018-08-24
Regression: no longer reading an extra byte.

timuralp:
Fixes a regression where we stopped reading an extra byte from Swift.
This results in ChunkWriteTimeout errors when retrieving objects from
Swift to use with 1space (affects the swift-s3-sync daemon).


timuralp:
@mdegerne unfortunately, the regression is in Swift and the patch I added to Swift that I thought fixes this issue, did not. At some point we should look more at why we hit ChunkWriteTimeout even if we read all of the bytes (according to Content-Length).

tipabu:
Maybe related? https://github.com/openstack/swift/commit/4a0afa9

This also reminds me that I should finish reviewing https://review.openstack.org/#/c/575254/ ...

PR #162, opened 2018-08-25
Update container crawler for parallel enumerators.

timuralp:
Pulls in the update to take advantage of parallel enumeration. Because
of the keystone lazy importer, we have to import keystone in
sync_swift.py, even though we are not actually going to use it
(otherwise, it will fail when we attempt to make requests against a
keystone endpoint).


timuralp:
I had to fix up one thing in 1space to make use of the new container crawler (namely, the keystone import). You can try the integration tests/play with it to try out the new parallel enumerators.

PR #165, opened 2018-08-29
Remote Delete After

mdegerne:
Add a new configuration option (remote_delete_after), currently only when
remote is swift, to set the x-delete-after header. It is intended as a lifecyle
option, but use with caution as data will be deleted from the remote store
when remote_delete_after seconds have elapsed.

Fixes https://github.com/swiftstack/1space/issues/172


test/integration/test_s3_sync.py
```diff
@@ -640,6 +640,41 @@ def test_swift_archive_range_get(self):
         clear_swift_container(self.swift_dst, mapping['aws_bucket'])
         clear_swift_container(self.swift_src, mapping['container'])
 
+    def test_swift_archive_slo_w_rem_del_at(self):
```
timuralp: I think we could spell out the test names. It will make it easier when they fail.

test/integration/test_s3_sync.py
```diff
@@ -640,6 +640,41 @@ def test_swift_archive_range_get(self):
...
+                return False
+
+        hdrs = wait_for_condition(5, _check_synced)
+        self.assertIn('x-delete-at', hdrs.keys())
+        target_time = int(time.time() + mapping['remote_delete_after'])
+        self.assertTrue(int(hdrs['x-delete-at']) <= target_time)
+        seg_container = mapping['aws_bucket'] + '_segments'
+        hdrs = self.remote_swift('head_object', seg_container, 'part1')
+        self.assertIn('x-delete-at', hdrs.keys())
+        self.assertTrue(int(hdrs['x-delete-at']) <= target_time)
```
timuralp: I'm a little worried about this `<=` condition. For example, we wouldn't catch a bug where all objects are set to be deleted immediately. If we could test the range, that would be better. One way to do it would be to call `time.time()` before `put_object` and then ensure the `x-delete-at` value is between that time + configured time and the target time.

PR #168, opened 2018-09-06
Fix a flaky test.

timuralp:
Fixes a race where an object may be in both locations, but not yet
expired. For the purposes of the test, it doesn't matter.


timuralp:
This has been a bit irritating lately. Merging this in.

PR #182, opened 2018-09-14
Reserve an S3 connection before uploading parts.

timuralp:
Fixes a regression where we stopped reserving the S3 connection before
making a Swift GET request. If we make the Swift request first, the
connection may timeout before we acquire a socket connected to S3.


timuralp:
@dbishop this is a regression from the cloud connector work. I added a test that would hopefully catch this in the future. I toyed with the idea of making the `_upload_part()` method either create the `FileWrapper` instance itself or pass in an optional `s3_client` argument, but wasn't particularly happy with either one of those. Let me know what you think. I prefer this as the smallest fix for currently broken multipart S3 uploads.

dbishop:
lgtm, from a cursory reading of the diff

timuralp:
Follow on commit fixes the cloud connector build.

PR #183, opened 2018-09-18
Shunt SLO object when unable to restore due to missing manifest

### LINKED ISSUE #173: If restore_object is set, GET hangs on S3 multipart-uploaded object without a manifest
If an object is put into an S3 through a multipart upload, with `restore_object` being set, we cannot fetch the object if the Swift manifest (`*.swift_slo_manifest`) does not exist. We should fail gracefully in this case and not attempt to restore the object, but still serve it back to the client.

thiagodasilva:
Previously if SLO manifest was missing from s3 endpoint
and restore_object was set. A GET would fail to restore object and
also fail to shunt object back to client.

This patch allows the shunt to continue while logging the
restore_object failure.

Fixes #173


s3_sync/shunt.py
```diff
@@ -419,10 +419,16 @@ def handle_object(self, req, start_response, sync_profile, obj,
...
+                if check_slo(put_headers):
+                    if manifest:
+                        app_iter = SwiftSloPutWrapper(
+                            app_iter, put_headers, req.environ['PATH_INFO'],
+                            self.app, manifest, self.logger)
+                    else:
+                        # if slo manifest is missing, log error, don't attempt
+                        # to restore object, but continue shunt
+                        self.logger.error('Failed to restore slo object due '
+                                          'to missing manifest: %s' % obj)
```
mdegerne: I don't see how this is working as you expect. Will it not just return the 404? There is no call in this path to `provider.shunt_object`.

s3_sync/utils.py
```diff
@@ -874,7 +874,8 @@ def convert_to_local_headers(headers, remove_timestamp=True):
     # restoring from Swift.
     if 'x-timestamp' in put_headers and remove_timestamp:
         del put_headers['x-timestamp']
-    if put_headers['etag'].startswith('"%s' % PFS_ETAG_PREFIX):
+    if put_headers.get('etag') and \
```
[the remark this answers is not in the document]
timuralp: I would prefer not making this change unless we have to. We don't do this today, but 1space should include the ETag/Content-MD5 when putting an object into Swift.

s3_sync/shunt.py
timuralp: That happens on line 418.

mdegerne: Thank you.

test/unit/test_shunt.py
```diff
@@ -604,6 +604,84 @@ def test_tee(self, mock_s3_shunt, mock_swift_shunt, mock_s3_get_manifest,
...
+             for e in self.swift.calls],
+            [
+                ('HEAD', '/v1/%s' % account),
+                ('GET', '/v1/%s/foo' % path),
+            ])
+        self.assertEqual(payload, resp_body)
+
+    @mock.patch.object(sync_s3.SyncS3, 'get_manifest')
+    @mock.patch.object(sync_s3.SyncS3, 'shunt_object')
+    def test_missing_slo_manifest2(
```
thiagodasilva: left by mistake, will fix

s3_sync/utils.py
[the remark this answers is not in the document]
thiagodasilva: I ran into this error while testing against s3api, will investigate again.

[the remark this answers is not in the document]
thiagodasilva: Ok, the reason I ran into this error is because I had setup my "target" to also be the same swift cluster (w/ s3api enabled), which means that for the GET of the manifest that we store on the target container, this code gets executed. The better fix was to actually move the call to convert_to_local_headers (shunt.py line 419) to inside the if block

PR #184, opened 2018-09-19
Safer Remote Delete After for SLOs

mdegerne:
This sets the delete after for SLO segments to be 24 hours later than
configured for objects. Since the segments are uploaded just before the
manifest, this ensures that the manifest expires before the objects.

### DIFF
--- a/s3_sync/sync_swift.py
+++ b/s3_sync/sync_swift.py
@@ -37,6 +37,8 @@ try:
 except ImportError:
     pass
 
+DEFAULT_SEGMENT_DELAY = 60 * 60 * 24
+
 
 class SyncSwift(BaseSync):
     def __init__(self, *args, **kwargs):
@@ -44,6 +46,10 @@ class SyncSwift(BaseSync):
         # Used to verify the remote container in case of per_account uploads
         self.verified_container = False
         self.remote_delete_after = self.settings.get('remote_delete_after', 0)
+        # Remote delete after addition for segments (defaults to 24 hours)
+        self.remote_delete_after_addition = \
+            self.settings.get('remote_delete_after_addition',
+                              DEFAULT_SEGMENT_DELAY)
 
     @property
     def remote_container(self):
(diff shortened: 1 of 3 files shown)


mdegerne:
The run_test change was because it was taking 42 seconds for swift-s3-sync to respond as desired.

s3_sync/sync_swift.py
```diff
@@ -527,8 +527,10 @@ def _upload_segment(self, segment, req_headers, internal_client):
                 self.account + segment['name'], segment['bytes']))
             seg_headers = self._client_headers()
             if self.remote_delete_after:
+                # Delete segments 1 day after manifest
+                remote_delete_after = self.remote_delete_after + (60 * 60 * 24)
```
timuralp: We may want this to be a configurable value. I can imagine in certain compliance situations with large objects, someone may want to set this higher (e.g. 5TB over a 100Mbps link would take 4.6 days). Could you plumb this through the configuration?

s3_sync/sync_swift.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, *args, **kwargs):
         # Used to verify the remote container in case of per_account uploads
         self.verified_container = False
         self.remote_delete_after = self.settings.get('remote_delete_after', 0)
+        # Remote delete after addition for segments (defaults to 24 hours)
+        self.remote_delete_after_addition = \
+            self.settings.get('remote_delete_after_addition', 60 * 60 * 24)
```
timuralp: Please define a constant for the default.

s3_sync/sync_swift.py
```diff
@@ -527,8 +530,11 @@ def _upload_segment(self, segment, req_headers, internal_client):
                 self.account + segment['name'], segment['bytes']))
             seg_headers = self._client_headers()
             if self.remote_delete_after:
+                # Delete segments 1 day after manifest
```
timuralp: Nit: This comment may not always be accurate.

PR #187, opened 2018-09-24
Preserve the segments container if per-account.

timuralp:
When we are copying a whole account (per-account is True in SyncSwift
and the mapping's container is set to /*), we should not rewrite the
destination segment containers. This resolves an issue with duplicating
storage when we copy the SLO, placing segments in container
<container>_segments, and then copying the segments container into
<segments_container>.


timuralp:
@mdegerne good point! At least we don't double the space usage, but extra transfers are definitely not great. I'll see if I can improve on that.

timuralp:
@mdegerne I made a follow-on commit to check the SLO segments. Please check it out.

PR #189, opened 2018-10-01
Add syslog capabilities to s3-sync

ctennis:


s3_sync/daemon_utils.py
```diff
@@ -42,14 +43,42 @@ def setup_logger(logger_name, config):
                                                        backupCount=5)
     else:
         raise RuntimeError('log file must be set')
+
     handler.setFormatter(formatter)
     logger.addHandler(handler)
 
-    for logname in ['boto3', 'botocore', 's3_sync']:
+    for logname in ['boto3', 'botocore']:
```
timuralp: I think the fix here should be to use `base_sync.LOGGER_NAME` here. I don't believe there is any situation where we don't want to configure that logger as part of the setup. We could check if `logger_name == base_sync.LOGGER_NAME` and include it in the list if that's not true.

I think we could improve this code a little bit by having the fourth item in the list be `logger_name` and removing the call to `logger.addHandler` on line 48 and the `setLevel` call on line 35.

s3_sync/daemon_utils.py
```diff
@@ -42,14 +43,42 @@ def setup_logger(logger_name, config):
...
+        host = syslog_config.get('host', 'localhost')
+        port = syslog_config.get('port', 514)
+        proto = syslog_config.get('proto', 'udp')
+
+        if proto == 'udp':
+            socktype = socket.SOCK_DGRAM
+        else:
+            socktype = socket.SOCK_STREAM
+
+        logger.debug("Using syslog")
```
timuralp: Looks like this log line will be emitted through the last logger in the `for`-loop on line 50. I think we'll want to be careful and use the logger we setup on line 34.

s3_sync/daemon_utils.py
```diff
@@ -42,14 +43,42 @@ def setup_logger(logger_name, config):
...
+        syslog_handler = logging.handlers.SysLogHandler(
+            address=(host, port),
+            socktype=socktype
+        )
+
+        syslog_handler.setFormatter(formatter)
+        logger.setLevel(level)
```
timuralp: Do we need to set the log level twice?

s3_sync/daemon_utils.py
```diff
@@ -42,14 +43,42 @@ def setup_logger(logger_name, config):
...
+        logger.addHandler(syslog_handler)
+
+        for logname in ['boto3', 'botocore']:
```
timuralp: Consider creating the `syslog_handler` before the loop on line 50 and iterating only once to add the handler.

s3_sync/migrator.py
```diff
@@ -1116,12 +1117,12 @@ def main():
...
 
-    handler = setup_logger(LOGGER_NAME, migrator_conf)
-    logger = logging.getLogger('s3_sync')
-    logger.addHandler(handler)
-
+    setup_logger(LOGGER_NAME, migrator_conf)
     load_swift(LOGGER_NAME, args.once)
 
+    # Also sure we have the s3-sync logger initialized
+    setup_logger(S3_SYNC_LOGGER_NAME, migrator_conf)
```
timuralp: I think we should always configure the `base_sync.LOGGER_NAME` in the `setup_logger` function. If we do, we do not need this call (it also duplicates the setup of the other loggers otherwise).

test/unit/test_daemon_utils.py
```diff
@@ -96,6 +98,19 @@ def test_log_file_handler(self, logging_mock):
             mock.call().setLevel('INFO'),
             mock.call().addHandler(handler)])
 
+    @mock.patch('s3_sync.daemon_utils.logging')
+    def test_log_file_handler(self, logging_mock):
```
timuralp: This is really `test_setup_logger`?

timuralp:
I suspect these failures are related to the syslog configuration. Containers are not reused on Travis, but we cache the images -- that shouldn't cause issues. Have you been able to reproduce this locally? I can reproduce this issue on your branch by running `./run_tests`.

ctennis:
ok Timur, I think specs are passing happily now

timuralp:
@ctennis great! Looks good to me. Do you mind squashing the commits before merging onto master?

PR #190, opened 2018-10-02
Add the statsd plumbing for copied objects.

timuralp:
Adds the metric for objects copied by swift-s3-sync. There is follow-on work where we want to distinguish PUT vs already existing objects, but deferring that to a subsequent patch. We also want to similarly expose
the number of objects copied, but that is similarly deferred to a follow-on patch.


s3_sync/sync_container.py
```diff
@@ -38,16 +39,31 @@ class SyncContainer(container_crawler.base_sync.BaseSync):
...
         self.copy_after = int(sync_settings.get('copy_after', 0))
         self.retain_local = sync_settings.get('retain_local', True)
         self.propagate_delete = sync_settings.get('propagate_delete', True)
+        self._settings = sync_settings
         self.provider = create_provider(sync_settings, max_conns,
                                         per_account=self._per_account)
+        self.statsd_client = statsd_client
+
+    @property
+    def statsd_prefix(self):
```
dbishop: Unless settings are mutable during the lifetime of an instance of this class, couldn't this string just get computed once and stored in `self` in this class' constructor?

s3_sync/sync_container.py
```diff
@@ -100,10 +116,17 @@ def save_last_row(self, row, db_id):
             json.dump(status, f)
             f.truncate()
 
+    def statsd(self, metric, value):
```
dbishop: wrapping the statsd_client's `increment` method with a method whose name drops the "increment" part seems bad; surely we're going to want to do timings and whatnot later?  Calling this `increment()` or `statsd_increment()` or something would be better.

s3_sync/sync_container.py
```diff
@@ -38,16 +39,31 @@ class SyncContainer(container_crawler.base_sync.BaseSync):
...
+    @property
+    def statsd_prefix(self):
+        parts = [
+            self._settings.get('aws_endpoint', 'S3'),
+            # NOTE: the custom_prefix is not included in the mapping
+            # identification
+            self.aws_bucket,
+            self._settings['account'],
+            self._settings['container']
+        ]
```
dbishop: Is not having `custom_prefix` in here somewhere going to cause or allow metric name collisions between distinct mappings?

timuralp: Today, we can't have a single account/container mapped to two different buckets, so there would be no conflicts. I think you're right that we should include `custom_prefix` for completeness and future-proofing this. Do you know if the following metric would be acceptable: `foo..bar` (notice the two ".")? Because sometimes the `custom_prefix` will be empty and I'm not sure of the best way to express it. Ideas?

s3_sync/sync_container.py
timuralp: Done.

s3_sync/sync_container.py
timuralp: Good point -- done.

timuralp:
@dbishop I think I addressed your comments. Let me know what you think!

PR #191, opened 2018-10-05
Changes for the updated ContainerCrawler library.

timuralp:
Adds the get/save_last_processed_row() and get/save_last_verified_row()
methods to the SyncContainer class.


s3_sync/sync_container.py
```diff
@@ -70,38 +73,65 @@ def get_last_row(self, db_id):
...
+        for field in self.POLICY_FIELDS:
+            policy[field] = getattr(self, field)
+
         if not os.path.exists(self._status_account_dir):
             os.mkdir(self._status_account_dir)
         if not os.path.exists(self._status_file):
             with open(self._status_file, 'w') as f:
-                json.dump({db_id: dict(last_row=row,
-                                       aws_bucket=self.aws_bucket)}, f)
+                json.dump({db_id: {row_field: row,
```
timuralp: I realized we have to be careful here and write down 0 for the last verified row, as otherwise we will pick up the last_row we just wrote down.

PR #192, opened 2018-10-05
Harden tests by checking for errors in the logger.

timuralp:
We have a number of success test failures where we weren't checking the
expected errors in the log or that errors were not logged.

This removes the silent failures to update object metadata, as well as
hardens existing tests.


timuralp:
Noticed the extraneous output from our tests and realized that a number of them weren't testing the things we thought they should be testing because of silent failures.

timuralp:
Merging this to not let it linger. It's all test hardening, so...

PR #194, opened 2018-10-10
Support migration from s3 endpoint

thiagodasilva:
Add basic support for s3 migration, migrate most objects
with the exception of MPU objects. It also added support to propagate
POSTs back to S3 if object has not been migrated yet.

Let to do in follow-on patches:
 - MPU/Large object support
 - list buckets
 - propagate deletes


s3_sync/sync_s3.py
```diff
@@ -308,13 +317,18 @@ def list_buckets(self, marker, limit, prefix, parse_time=True):
...
            so that parameter is ignored, as well.
         '''
-        resp = self._call_boto('list_buckets')
-        resp.body = [
-            {'last_modified': bucket['CreationDate'],
-             'count': 0,
-             'bytes': 0,
-             'name': bucket['Name']} for bucket in resp.body['Buckets']]
-        return resp
+        with self.client_pool.get_client() as s3_client:
```
timuralp: It would be nice to keep moving toward using the `_call_boto` helper. What blows up when we try to use it?

s3_sync/utils.py
```diff
@@ -931,7 +931,10 @@ def _results_iterator(_resp):
                     marker = item['name']
                 else:
                     marker = item['subdir']
-                item['content_location'] = [item['content_location']]
+
+                if 'content_location' in item:
+                    item['content_location'] = [item['content_location']]
```
timuralp: Looks like this comes up because S3 bucket listings do not include content_location, but we do populate it for Swift listings. Why do we want the SyncS3 and SyncSwift providers to return different results in that case?

s3_sync/sync_s3.py
```diff
@@ -308,13 +317,18 @@ def list_buckets(self, marker, limit, prefix, parse_time=True):
...
+            resp = s3_client.list_buckets()
+            response_body = [
+                {'last_modified': bucket['CreationDate'],
+                 'count': 0,
+                 'bytes': 0,
+                 'name': bucket['Name']} for bucket in resp.get('Buckets')]
```
timuralp: I think we want to include `content_location`.

s3_sync/sync_s3.py
thiagodasilva: `_call_boto` doesn't handle `Buckets` in the response, only `Body`. I could have added the support there, but prefered to follow the same path as list_objects. I'm not convinced one is better than the other. Thoughts?

timuralp: I thought we're pushing this into a follow-on patch? Let's defer the discussion to that PR. The reason that the `_call_boto()` approach is a bit cleaner is because we have one place to do any request mutation of headers and one place to mutate the responses into a common form.

PR #195, opened 2018-10-12
Add a SyncContainerFactory class.

timuralp:
Allows us to work against a new version of the container-crawler
library, using a factory pattern.


timuralp:
This requires a new version of ContainerCrawler to not fail.

s3_sync/sync_container.py
```diff
@@ -162,3 +162,14 @@ def handle(self, row, swift_client):
                 except UnexpectedResponse as e:
                     if '409 Conflict' in e.message:
                         pass
+
+
+class SyncContainerFactory(object):
```
tipabu: Should we define a friendlier `__str__` for this? Looks like we used to log something like
```
Failed to process account/container with SyncContainer
```
but currently it'd be something like
```
Failed to process account/container with <s3_sync.sync_container.SyncContainerFactory object at 0x108861b10>
```
?

timuralp: Done.

PR #196, opened 2018-10-12
Ensure internal requests are closed on failures.

timuralp:
We have to make sure to close the InternalClient request in case a
request to the remote provider fails. Otherwise, a ChunkWriteTimeout
error may be raised at some point in the green thread, leading to
subsequent failures


timuralp:
Agreed, I don't think we ever call `SLOFileWrapper.next`.

PR #197, opened 2018-10-13
Don't assume iterables are iterators

tipabu:
Following https://github.com/boto/botocore/commit/ec9f7bd (I think), I started seeing test failures complaining that

    TypeError: StreamingBody object is not an iterator

This is accurate: `StreamingBody` defined a `__iter__()` method (making it an iterable), but neither a `next()` method nor a `__next__()` method (which are required to be an iterator under python 2 or 3, respectively).


tipabu:
Submitted https://github.com/boto/botocore/pull/1580 FWIW.

PR #199, opened 2018-10-13
Skip MPU and large files

mdegerne:
We don't currently support migration from S3 Multipart Uploads or files larger
than the max_file_size setting. This skips these types of files and writes a
warning in the log.


test/integration/test_daemon_utils.py
```diff
@@ -7,6 +8,7 @@ def test_syslog(self):
         initialize_loggers({'log_file': '/tmp/test_file',
                             'log_level': 'debug',
                             'syslog': {'host': 'swift-s3-sync'}})
+        time.sleep(1)
```
tipabu: ??

Is this actually necessary, or was it just for debugging?

s3_sync/migrator.py
```diff
@@ -57,6 +58,7 @@
 MigrateObjectWork = namedtuple('MigrateObjectWork', 'aws_bucket container key')
 UploadObjectWork = namedtuple('UploadObjectWork', 'container key object '
                               'headers aws_bucket')
+S3_MPU_RE = re.compile('[0-9a-z]+-\d+$')
```
tipabu: Technically we could probably limit it to `[0-9a-f]{32}-\d+` but w/e.

s3_sync/migrator.py
```diff
@@ -853,6 +855,23 @@ def _find_missing_objects(
             elif local['name'] < remote['name']:
                 self._reconcile_deleted_objects(container, local['name'])
                 local = next(local_iter)
+            elif S3_MPU_RE.match(remote['hash']):
```
timuralp: This `elif` doesn't make sense. We should either make the check within the if clause on line 846 or in the `migrate_object` method. As is, we will try to migrate the object on line 848 anyway.

test/integration/test_daemon_utils.py
mdegerne: This was needed for my tests to pass. Presumably, the log hadn't written out yet.

s3_sync/migrator.py
mdegerne: Specifically didn't because the tests cheat on etag values.

s3_sync/migrator.py
mdegerne: You're right. I've changed it.

s3_sync/migrator.py
```diff
@@ -853,6 +870,11 @@ def _find_missing_objects(
             elif local['name'] < remote['name']:
                 self._reconcile_deleted_objects(container, local['name'])
                 local = next(local_iter)
+            elif self._is_large(remote):
+                scanned += 1
+                remote = next(source_iter)
```
timuralp: This `elif` is still confusing -- we make the `_old_enough` check in the `else` clause below. I'd suggest performing the check in `_migrate_object`, which would make the logic of increasing the iterators clear (otherwise, notice how the local iterator is not incremented and should be.

timuralp: Ok, we can't move it into `_migrate_object` because we lose the ETag and Last-Modified date. I'd either move this conditional next to the other `old_enough` check or make a new method that enqueues the work (calls `self.object_queue.put()`) and does all the checking beforehand. Either way, please remove this `elif`, as we are checking here for the object names and their ordering and I would strongly prefer to leave these outer conditionals only related to that.

mdegerne: I've implemented as suggested. I think that we should keep this little function anyway as a place to drop in the metadata check - if we implement it for migrations.

PR #201, opened 2018-10-16
Support Non-printable ASCII characters in AWS S3 metadata

thiagodasilva:
When syncing metadata to s3, encode non-ascii characters
and when migrating metadata from s3, make sure to decode.

[Fixes #156776841]


s3_sync/utils.py
```diff
@@ -845,6 +848,24 @@ def convert_to_s3_headers(swift_headers):
...
+    if value.startswith('=?utf-8?q?'):
+        return '=?UTF-8?Q?' + value[10:]
+    elif value.startswith('=?utf-8?b?'):
+        return '=?UTF-8?B?' + value[10:]
+    else:
+        return value
+
+
+def decode_s3_header(value):
+    header, charset = decode_header(value)[0]
```
timuralp: Should we check the returned `charset` value?

s3_sync/utils.py
```diff
@@ -845,6 +848,27 @@ def convert_to_s3_headers(swift_headers):
...
+    if charset not in ('utf-8', 'UTF-8', 'utf8'):
```
mdegerne: Out of curiosity, how did you come up with this specific list of permutations?

thiagodasilva: TBH just out of experience in our own code, not a very empirical method. utf8 should actually not be used according to standards[1], but I see it used all the time :man_shrugging: 

[1] - https://en.wikipedia.org/wiki/UTF-8#Official_name_and_variants

s3_sync/utils.py
thiagodasilva: done, added the check

s3_sync/utils.py
```diff
@@ -854,15 +878,15 @@ def convert_to_swift_headers(s3_headers):
             swift_headers[header[len(S3_USER_META_PREFIX):]] = value
         elif header.startswith(S3_USER_META_PREFIX):
             key = SWIFT_USER_META_PREFIX + header[len(S3_USER_META_PREFIX):]
-            swift_headers[key] = value
+            swift_headers[key] = decode_s3_header(value)
```
timuralp: Please add unit tests to cover the mime-encoding conversion.

s3_sync/utils.py
```diff
@@ -845,6 +848,27 @@ def convert_to_s3_headers(swift_headers):
     return s3_headers
 
 
+def sanitize_s3_header(value):
```
timuralp: Can you add unit tests for this function and/or expand the unit tests for `convert_to_s3_headers()`?

s3_sync/utils.py
```diff
@@ -845,6 +848,27 @@ def convert_to_s3_headers(swift_headers):
...
+        return '=?UTF-8?B?' + value[10:]
+    else:
+        return value
+
+
+def decode_s3_header(value):
+    header, charset = decode_header(value)[0]
+    if charset not in ('utf-8', 'UTF-8', 'utf8'):
+        return value
+    else:
```
timuralp: Can drop the else and just return header.

PR #202, opened 2018-10-17
Update 1space to use new boto3 and botocore

thiagodasilva:
New version of botocore carries only one of our own
patches (i.e., S3SigV4ChunkedAuth)


requirements.txt
```diff
@@ -1,5 +1,5 @@
-boto3==1.3.1
-git+https://github.com/swiftstack/botocore.git@1.4.32.6#egg=botocore
+boto3==1.9.23
+git+https://github.com/swiftstack/botocore.git@ss-1.12.23#egg=botocore
```
timuralp: I think we should still create a tag like "1.12.23.1" because when we build the package, we need a new version between the botocore releases and that version will be used as the deb/rpm version as well.

test/integration/test_cloud_connector.py
```diff
@@ -77,8 +79,8 @@ def test_container_get_just_s3_objs(self):
                 obj_name, {'content-length': str(len(obj_name) + 1),
                            'content-type': 'application/party-time'},
                 body_iter)
-            exp_etags[obj_name] = '"' + \
-                hashlib.md5(''.join(body_iter)).hexdigest() + '"'
+            etag = '"' + hashlib.md5(''.join(body_iter)).hexdigest() + '"'
+            exp_etags[obj_name] = urllib.quote(etag)
```
timuralp: This is surprising to me. The new version of botocore returns an S3 tag deadbeef as %22deadbeef%22 and not "deadbeef"?

thiagodasilva: Yeah, it was surprising to me too, but it is not only the new version of botocore, but it also only happens when targeting swift3 (or s3api). When the endpoint is either aws or s3proxy, then etag is not quoted. Note below that `NextContinuationToken` is also quoted

timuralp: Oh, interesting. Is there a bug in swift3/s3api? @tipabu do you know anything about this?

timuralp: ETags with %22 would violate the ETag RFC even more than the ones that lack quotes (I think). RFC: https://tools.ietf.org/html/rfc7232#section-2.3

tipabu: Submitted https://review.openstack.org/#/c/611721/ but I may have forgotten some places. There's a decided lack of coverage for this feature 🙁

thiagodasilva: @tipabu will review swift patch. @timuralp agree we will need to support decoding for a while.

PR #203, opened 2018-10-26
Switch to using UNSIGNED-PAYLOAD with AWS S3.

timuralp:
We don't have to pay the overhead of chunked encoding by switching to
UNSIGNED-PAYLOAD.


timuralp:
We need to make sure everything works as expected against real S3, but I didn't encounter problems in my limited testing.

thiagodasilva:
code looks good, automated tests pass and manual tests to s3 also worked without issues.

PR #205, opened 2018-10-30
Remove X-Newest header from 1space requests.

timuralp:
X-Newest, while improving consistency guarantees, can be problematic. In the case of EC, using this header can cause a 10x increase in HEAD requests, as there would be one HEAD request per part.


timuralp:
It's probably best to look at the commits and only review the X-Newest commit specifically. The other commit is a prerequisite for this PR.

PR #207, opened 2018-10-31
Supply ContentMD5 when uploading to S3.

timuralp:
Supplies the object's ETag when uploading to S3 to ensure integrity of
the objects transferred from Swift.


timuralp:
We should've done this earlier, but better late than never? @tipabu I think this takes care of the concern of writing corrupt objects into S3?

PR #208, opened 2018-11-01
Migrate large s3

mdegerne:


s3_sync/migrator.py
```diff
@@ -1008,29 +1141,44 @@ def _migrate_slo(self, aws_bucket, slo_container, key, resp):
         except eventlet.queue.Full:
             self._upload_object(work)
 
-    def _upload_object(self, work):
+    def _upload_object(self, work, migrated_object=True, use_put=False):
```
thiagodasilva: maybe add some docs to explain args

s3_sync/migrator.py
```diff
@@ -935,16 +950,134 @@ def _migrate_object(self, aws_bucket, container, key):
...
+                part_resp.body.close()
+                raise MigrationError('Failed to GET "%s/%s(part %d) %d": %s'
+                                     % (aws_bucket, key, part + 1,
+                                        part_resp.status,
+                                        part_resp.headers))
             put_headers = convert_to_local_headers(
-                resp.headers.items(), remove_timestamp=False)
+                part_resp.headers.items(), remove_timestamp=False)
+            sz_bytes = int(put_headers['Content-Length'])
+            del(put_headers['etag'])
```
thiagodasilva: why delete the etag header? could be used by swift for verification, no?

s3_sync/migrator.py
```diff
@@ -935,16 +950,134 @@ def _migrate_object(self, aws_bucket, container, key):
...
-                container, key, FileLikeIter(resp.body), put_headers,
-                aws_bucket)
-            self._upload_object(work)
+                segment_container, segment_prefix + segment_key,
+                FileLikeIter(part_resp.body), put_headers, aws_bucket)
+            put_resp = self._upload_object(
+                work, migrated_object=False, use_put=True)
+            new_seg = {'name': '/' + segment_container + '/' + segment_prefix +
+                       segment_key, 'bytes': sz_bytes,
+                       'hash': put_resp.headers['etag']}
```
thiagodasilva: could have an assert that the put_resp etag is the same as the put_headers etag

test/integration/migrator_utils.py
```diff
@@ -39,4 +39,5 @@ def get_migrator(self, migration, status):
             self.config, self.SWIFT_DIR, self.migrator_config['workers'] + 1)
         return s3_sync.migrator.Migrator(
             migration, status, self.migrator_config['items_chunk'],
-            self.migrator_config['workers'], ic_pool, self.logger, 0, 1)
+            self.migrator_config['workers'], ic_pool, self.logger, 0, 1,
+            self.migrator_config.get('segment_size', 100000000))
```
thiagodasilva: for testing purposes, i'd suggest using smaller segment_size

test/integration/test_migrator.py
```diff
@@ -1469,3 +1473,104 @@ def _check_migrated_objects(container, expected_count):
...
+        return client.complete_multipart_upload(
+            Bucket=self.aws_bucket,
+            Key=key,
+            UploadId=upload_id,
+            MultipartUpload={
+                'Parts': [{'PartNumber': i + 1, 'ETag': parts_etags[i]}
+                          for i in range(parts)]
+            })
+
+    def test_s3_mpu_migration(self):
```
thiagodasilva: how about a test for creating SLOs from large non-mpu objects?

run_tests
```diff
@@ -38,5 +38,5 @@ docker exec swift-s3-sync timeout 40 \
              --os-project-domain-name default --os-user-domain-name default \
              --auth-version 3 --debug stat; do sleep 0.5; done' >/dev/null 2>&1
 
-docker exec -e DOCKER=true  swift-s3-sync nosetests /swift-s3-sync/test/integration/
+docker exec -e DOCKER=true -e AWS_IDENTITY=$AWS_IDENTITY -e AWS_SECRET=$AWS_SECRET -e AWS_BUCKET=$AWS_BUCKET  swift-s3-sync nosetests /swift-s3-sync/test/integration/
```
thiagodasilva: maybe add a line above explaining what the env. variables are used for

s3_sync/migrator.py
```diff
@@ -839,8 +854,8 @@ def _is_large(self, entry):
         return False
 
     def object_queue_put(self, aws_bucket, container, remote):
-        if self._is_large(remote):
-            return
+        # if self._is_large(remote):
```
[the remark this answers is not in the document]
mdegerne: Yep. Meant to.

s3_sync/migrator.py
```diff
@@ -935,16 +950,134 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        remote_etag = headers['etag']
+        del(headers['etag'])
+        nparts = nparts_from_etag(remote_etag)
+        segments = []
+        ts = _create_x_timestamp_from_hdrs(headers)
+        segment_container = "%s_segments" % (container,)
+        content_length = int(resp.headers['Content-Length'])
+        segment_size = self.segment_size
+        segment_prefix = "%s/%s/%s/%s/" % (
+            key, ts, content_length, segment_size)
```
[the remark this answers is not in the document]
mdegerne: You are right. I don't have part_size until later, I'll move this.

test/integration/test_migrator.py
```diff
@@ -1469,3 +1473,104 @@ def _check_migrated_objects(container, expected_count):
...
+        # Create mpu on S3
+        key = 'mpu_test'
```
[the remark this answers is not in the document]
mdegerne: Agreed. I think I ran into some issues with testing/deleting/re-testing due to this. I like the idea of UUID.

s3_sync/migrator.py
```diff
@@ -935,16 +950,134 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        for part in range(nparts):
+            segment_key = "%08d" % (part + 1,)
+            args = {'bucket': aws_bucket, 'PartNumber': part + 1}
+            part_resp = self.provider.get_object(key, **args)
+            if part_resp.status != 206:
+                resp.body.close()
```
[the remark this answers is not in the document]
mdegerne: Yes. I'll add that.

s3_sync/migrator.py
```diff
@@ -935,16 +950,134 @@ def _migrate_object(self, aws_bucket, container, key):
...
+                else:
+                    self.logger.error(
+                        'Failed to delete segment %s, name does not start '
+                        'with %s' % (seg, seg_container))
+                    continue
+                headers = {}
+                ic.delete_object(self.config['account'], seg_container, key,
+                                 headers)
+
+    def _migrate_mpu(self, aws_bucket, container, key, resp):
```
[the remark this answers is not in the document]
mdegerne: I think I want to defer trying to optimize for the transient failure case, but it might be a useful thing to do.

s3_sync/migrator.py
mdegerne: The etag header, unfortunately, is the same etag as the primary object (i.e. xxxxxx-N).

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+            return
+        put_headers = convert_to_local_headers(
+            resp.headers.items(), remove_timestamp=False)
+        content_length = int(resp.headers['Content-Length'])
+        if S3_MPU_RE.match(resp.headers.get('etag', '')):
+            self._migrate_mpu(aws_bucket, container, key, resp)
+            return
+        if content_length > self.segment_size:
+            self._create_slo(
+                container, key, resp.body, content_length,
```
timuralp: What if we pass `resp` here, similar to the large object handling, as opposed to the individual pieces?

s3_sync/migrator.py
mdegerne: The etag for the part is the same as the etag for the whole - useless in this case.

s3_sync/migrator.py
mdegerne: See above.

test/integration/migrator_utils.py
mdegerne: I am using a very large value here specifically to avoid having it change existing tests. For specific testing of the functionality, I am using a smaller value.

test/integration/test_migrator.py
mdegerne: Working on that today 😀

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+                put_headers, aws_bucket)
+            return
+        work = UploadObjectWork(
+            container, key, FileLikeIter(resp.body), put_headers, aws_bucket)
+        self._upload_object(work)
+
+    def _s3_etag_from_manifest(self, manifest):
```
timuralp: Can you replace this with `s3_sync.utils.get_slo_etag`?

run_tests
mdegerne: I was wondering the best place(s) to do this, and was hoping for suggestions. I'll do this at a minimum.

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        segment_container = "%s_segments" % (container,)
+        content_length = int(resp.headers['Content-Length'])
+        segment_size = self.segment_size
+        segment_prefix = "%s/%s/%s/%s/" % (
+            key, ts, content_length, segment_size)
+        for part in range(nparts):
+            segment_key = "%08d" % (part + 1,)
+            args = {'bucket': aws_bucket, 'PartNumber': part + 1}
+            part_resp = self.provider.get_object(key, **args)
+            if part_resp.status != 206:
```
timuralp: Alternatively, we could check `part_resp.success`. That may be more robust in case some S3 clone doesn't return 206.

s3_sync/migrator.py
timuralp: Please add a comment about why the ETag is not useful. Also, in this case, we may want to compute the MD5 while uploading the part, which would allow us to fail before uploading the entire object (and retry the failing part at some point when we add retries).

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+                                 (aws_bucket, key))
+        manifest = json.dumps(segments)
+        headers['Content-Length'] = len(manifest)
+        headers[REMOTE_ETAG] = remote_etag
+        headers['X-Static-Large-Object'] = 'True'
+        work = UploadObjectWork(container, key, FileLikeIter(manifest),
+                                headers, aws_bucket)
+        self._upload_object(work)
+
+    def _create_slo(self, container, key, data, content_length, headers,
```
timuralp: Nit: naming suggestion to keep in line with the other methods -- `_migrate_as_slo`. Also, consider passing the response, rather than the body, content-length, and headers separately.

s3_sync/migrator.py
```diff
@@ -935,16 +950,134 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        if 'etag' in headers:
+            remote_etag = headers['etag']
+            del(headers['etag'])
+        segments = []
+        segment_start = 0
+        segment_size = self.segment_size
+        ts = _create_x_timestamp_from_hdrs(headers)
+        segment_container = "%s_segments" % (container,)
+        segment_prefix = "%s/%s/%s/%s/" % (
+            key, ts, content_length, segment_size)
```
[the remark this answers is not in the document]
timuralp: I think we should use a naming convention as close to python-swiftclient as possible, which is:
```
2115                         segment_name = '%s/slo/%s/%s/%s/%08d' % (
2116                             obj, put_headers['x-object-meta-mtime'],
2117                             full_size, options['segment_size'], segment
```

The `segment` here is the part number/segment number. That's what Mandell is doing here. I think this is fine.

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+                content, segment_headers, aws_bucket)
+            put_resp = self._upload_object(
+                work, migrated_object=False, use_put=True)
+            # Add segment to segments list
+            new_seg = {'name': '/' + segment_container + '/' + segment_prefix +
+                       segment_key, 'bytes': segment_size,
+                       'hash': put_resp.headers['etag']}
+            segments.append(new_seg)
+            segment_start += segment_size
+        manifest = json.dumps(segments)
```
timuralp: Some of these lines are repeated between `_migrate_mpu` and `_create_slo`. Could we factor them out?

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        headers[REMOTE_ETAG] = remote_etag
+        headers['X-Static-Large-Object'] = 'True'
+        work = UploadObjectWork(container, key, FileLikeIter(manifest),
+                                headers, aws_bucket)
+        self._upload_object(work)
+
+    def _create_slo(self, container, key, data, content_length, headers,
+                    aws_bucket):
+        remote_etag = None
+        if 'etag' in headers:
```
timuralp: When would the `etag` not be in headers?

s3_sync/migrator.py
```diff
@@ -76,6 +80,13 @@ def __unicode__(self):
             self.container, self.account)
 
 
+def nparts_from_etag(etag):
+    m = re.match('[0-9a-z]+-(\d+)$', etag)
```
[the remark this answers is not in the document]
mdegerne: Yes, fixing.

s3_sync/migrator.py
```diff
@@ -726,6 +738,10 @@ def _check_large_objects(self, aws_bucket, container, key, client):
                     'Matching date, but differing SLO manifests'))
             return
 
+        if 'etag' in remote_resp.headers and REMOTE_ETAG in local_meta:
```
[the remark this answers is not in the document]
mdegerne: You're right. I had some intermediate code that sometimes removed it (was a bad idea). Fixing.

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+            return
+        if 'x-static-large-object' in resp.headers:
             # We have to move the segments and then move the manifest file
             self._migrate_slo(aws_bucket, container, key, resp)
-        else:
+            return
+        put_headers = convert_to_local_headers(
+            resp.headers.items(), remove_timestamp=False)
+        content_length = int(resp.headers['Content-Length'])
+        if S3_MPU_RE.match(resp.headers.get('etag', '')):
```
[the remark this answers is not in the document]
mdegerne: sure.

s3_sync/migrator.py
```diff
@@ -59,6 +60,8 @@
 UploadObjectWork = namedtuple('UploadObjectWork', 'container key object '
                               'headers aws_bucket')
 S3_MPU_RE = re.compile('[0-9a-z]+-\d+$')
+REMOTE_ETAG = get_object_transient_sysmeta(
```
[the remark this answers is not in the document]
mdegerne: The POST would make the local object newer, would it not? If so, the etag no longer matters.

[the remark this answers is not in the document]
mdegerne: @timuralp I'll move to s3_sync.utils.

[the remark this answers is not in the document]
timuralp: @mdegerne you're right! Once the POST happens, it will update the date on the object regardless and we wouldn't be checking if it's been synced or not.

s3_sync/migrator.py
```diff
@@ -1008,29 +1131,44 @@ def _migrate_slo(self, aws_bucket, slo_container, key, resp):
...
+            del headers['last-modified']
         size = int(headers['Content-Length'])
+        result = None
         with self.ic_pool.item() as ic:
             try:
-                ic.upload_object(
-                    content, self.config['account'], container, key,
-                    headers)
-                self.logger.debug('Copied "%s/%s"' % (container, key))
+                if use_put:
```
timuralp: I'm a little confused by this. The `InternalClient.upload_object()` does this:
```
752         headers = dict(headers or {})
753         if 'Content-Length' not in headers:
754             headers['Transfer-Encoding'] = 'chunked'
755         path = self.make_path(account, container, obj)
756         self.make_request('PUT', path, headers, (2,), fobj)
```

I suspect the problem is that it does not return anything, right? I think we could remove the `else` and always use this until upstream interface improves (not sure if there would be enthusiasm to expand this). Then we can clobber the `else` and the extra parameter.

s3_sync/migrator.py
timuralp: I think we can drop the `use_put` parameter -- see below. I would also rename the `migrated_object` parameter to be more explicit about what it's doing, e.g. `preserve_timestamp`.

s3_sync/utils.py
```diff
@@ -215,7 +215,8 @@ def _account_data_delivered(self, data):
         if self.length is not None:
             bytes_we_can_return = self.length - self.tell()
             if len(data) > bytes_we_can_return:
-                self.buf = None
+                # self.buf = None
```
timuralp: Dead code

s3_sync/migrator.py
mdegerne: Adding a comment. Not sure about how computing MD5 at this point helps significantly since I would assume that higher likelihood failure is coming from S3 in which case we wouldn't be able to catch until the end anyway. Best case, we can verify that swift ends up with the same MD5 that we calculate on the fly. I think this is an optimization that would best go into ic.upload_object -> when etag is not supplied verify that bits received match bits sent. Second best is to add an MD5 calculator into SeekableFileLikeIter somehow. Let me know if you feel it is worth working on now or if it can wait for next iteration.

s3_sync/utils.py
```diff
@@ -215,7 +215,8 @@ def _account_data_delivered(self, data):
...
+                self.buf = data[bytes_we_can_return:]
```
timuralp: I'm not convinced this is correct. Can you explain a little bit this change with respect to what `read()` and `next()` do, respectively?

s3_sync/migrator.py
mdegerne: already done.

s3_sync/migrator.py
mdegerne: Good idea.

s3_sync/utils.py
```diff
@@ -279,6 +280,14 @@ def __len__(self):
         raise TypeError("object of type '%s' has no len()" %
                         self.__class__.__name__)
 
+    def continue_reading(self, length=None):
```
timuralp: Can you explain a little bit why this is necessary? We should read the entire length of this stream, so I'm not sure why `_bytes_delivered` should be reset.

s3_sync/migrator.py
mdegerne: same comment as before - I had a version which stripped it out sometimes - was a bad idea. Fixing.

s3_sync/migrator.py
```diff
@@ -935,16 +951,123 @@ def _migrate_object(self, aws_bucket, container, key):
...
+            remote_etag = headers['etag']
+            del(headers['etag'])
+        segments = []
+        segment_start = 0
+        segment_size = self.segment_size
+        ts = _create_x_timestamp_from_hdrs(headers)
+        segment_container = "%s_segments" % (container,)
+        segment_prefix = "%s/%s/%s/%s/" % (
+            key, ts, content_length, segment_size)
+        content = SeekableFileLikeIter(data, length=segment_size)
```
timuralp: I have a few questions about this.
1. Why does this need to be seekable? Do we ever try to rewind this stream?
2. Why do we use the truncated length? Does InternalClient read past the boundary if we set the Content-Length?

I might be missing something here -- more context would be great!

s3_sync/migrator.py
```diff
@@ -971,14 +970,25 @@ def _migrate_object(self, aws_bucket, container, key):
...
-        etags = []
-        for part in manifest:
-            etags.append(part['hash'].decode('hex'))
-        return hashlib.md5("".join(etags)).hexdigest() + "-" + "%d" % (
-            len(etags))
+    def _delete_parts(self, seg_container, segmentlist):
+        with self.ic_pool.item() as ic:
+            for seg in segmentlist:
+                key = None
+                if seg['name'][1:len(seg_container) + 1] == seg_container:
```
timuralp: When do we have manifests that have entries not starting with "/"?

s3_sync/migrator.py
```diff
@@ -997,23 +980,25 @@ def _migrate_mpu(self, aws_bucket, container, key, resp):
...
-            key, ts, content_length, segment_size)
+        segment_prefix = None
         for part in range(nparts):
             segment_key = "%08d" % (part + 1,)
             args = {'bucket': aws_bucket, 'PartNumber': part + 1}
             part_resp = self.provider.get_object(key, **args)
             if part_resp.status != 206:
                 resp.body.close()
                 part_resp.body.close()
+                self._delete_parts(segments)
```
timuralp: The method expects two arguments. You probably noticed this already.

s3_sync/migrator.py
timuralp: Comment definitely helps. I think we can maybe defer the MD5 checking, as it's an optimization.

Here's my thinking: when using AWS S3 (I'll set aside S3 clones and HTTP endpoints for now), boto defaults to HTTPS, so we have some certainty that whatever bytes S3 produced we received faithfully. However, when writing to Swift, we use HTTP. The MD5 check would allow us to catch any transport issues in that path (where they might be more likely). We do validate this in the end by checking the merkle hash, so you're right that it's an optimization at best. I wouldn't spend time on this right now.

mdegerne:
Internal client "make_request" does a seek(0) on the body object it is
passed before sending, this is the source of most of the work needed on
SeekableFileLikeIter. Almost everything stems from that. If I don't set the
Content-Length to the segment length, swift complains if the stream ends
too soon.

mdegerne:
Re: manifest entries not starting with '/'. That was defensive coding for
the time I wrote the _delete_segments bit. Not actually needed.

timuralp:
@mdegerne I found the relevant bits in `make_request()`:
```
190                 if hasattr(body_file, 'seek'):
191                     body_file.seek(0)
```
What if we pass `seek_zero_cb` argument to `SeekableFileLikeIter`? It could be a NOOP and bypass that check.

s3_sync/migrator.py
```diff
@@ -935,16 +931,128 @@ def _migrate_object(self, aws_bucket, container, key):
...
+        nparts = nparts_from_etag(remote_etag)
+        segments = []
+        ts = _create_x_timestamp_from_hdrs(headers)
+        segment_container = "%s_segments" % (container,)
+        content_length = int(resp.headers['Content-Length'])
+        segment_prefix = None
+        for part in range(nparts):
+            segment_key = "%08d" % (part + 1,)
+            args = {'bucket': aws_bucket, 'PartNumber': part + 1}
+            part_resp = self.provider.get_object(key, **args)
```
tipabu: I was expecting this to have an `If-Match` header plumbed in, to make sure the object hasn't changed while we're uploading...

mdegerne:
That sounds important. I'll add that.

s3_sync/migrator.py
mdegerne: OK. Done.

s3_sync/migrator.py
mdegerne: done.

s3_sync/migrator.py
mdegerne: done as much as I can see.

s3_sync/migrator.py
mdegerne: Done.

s3_sync/utils.py
mdegerne: Thanks.

s3_sync/migrator.py
mdegerne: resolved.

s3_sync/migrator.py
mdegerne: fixed.

s3_sync/utils.py
mdegerne: Since the `length` is now the length of the content that can be `read()`, not the source content, we need to keep the data in the buffer that was read from source (vie `next()`) that extends beyond the current segment being written.

s3_sync/utils.py
mdegerne: _migrate_as_slo uses SeekableFileLikeIter to chunk out a single long stream as several smaller files. It initializes it with the length of the first segment. This allows the internal mechanism to keep track of how much data is left to deliver to the current reader. `continue_reading` re-initializes for the next segment.

s3_sync/migrator.py
```diff
@@ -1008,29 +1120,31 @@ def _migrate_slo(self, aws_bucket, slo_container, key, resp):
...
         size = int(headers['Content-Length'])
         with self.ic_pool.item() as ic:
             try:
-                ic.upload_object(
-                    content, self.config['account'], container, key,
-                    headers)
-                self.logger.debug('Copied "%s/%s"' % (container, key))
+                path = ic.make_path(self.config['account'], container, key)
+                result = ic.make_request(
+                    'PUT', path, dict(headers), (2,), content)
```
timuralp: What other type can `headers` be here?

mdegerne:
As discussed, added the reraise() call and a comment about `dict(headers)`

PR #211, opened 2018-11-06
Hardening and test improvements.

timuralp:
Broke this up into smaller commits -- let me know if separate PRs would be easier.


test/integration/utils.py
```diff
@@ -81,3 +88,76 @@ def get_messages(self):
...
+            settings,
+            per_account=per_account,
+            statsd_client=statsd_client)
+
+
+def get_container_crawler(profile, **kwargs):
+    defaults = {
+        'log_level': 'info',
+        'log_file': '/var/log/swift-s3-sync.log',
+        'verification_slack': 60,
```
thiagodasilva: don't we want this to be 0 for tests?

containers/swift-s3-sync/launch.sh
```diff
@@ -36,11 +36,6 @@ cd /swift-s3-sync; pip install -e .
 
 /usr/bin/sudo -u swift /swift/bin/startmain
 
-python -m s3_sync --log-level debug \
```
thiagodasilva: This is an interesting point to how we have been using the containers. 
On the README we tell readers to test 1space with these containers and as they upload objects they would see them being synced. This change here breaks that, so while I agree that the overall PR is a good change, it does expose the issue of trying to have one solution to demo 1space and to also run our CI tests. Thoughts?

PR #212, opened 2018-11-07
Do not require continue_reading.

timuralp:
Reworks the changes to SeekableFileLikeIter to reduce the interface
changes.


timuralp:
Tagged Darrell as the original author of the SeekableFileLikeIter.

s3_sync/migrator.py
```diff
@@ -1027,17 +1028,20 @@ def _migrate_as_slo(self, aws_bucket, container, key, resp):
...
             segment_headers['Content-Length'] = str(segment_size)
+            wrapped_data = SeekableFileLikeIter(
+                itertools.chain(buf, data), length=segment_size)
             new_seg = self._put_segment(
-                segment_container, segment_prefix, segment_key, content,
-                segment_size, segment_headers, aws_bucket)
+                segment_container, segment_prefix, segment_key,
+                wrapped_data, segment_size, segment_headers, aws_bucket)
+            if wrapped_data.buf:
+                buf = wrapped_data.buf
```
mdegerne: I think you may need to make this unconditional, so that you don't re-use buf if wrapped_data.buf is empty.

timuralp: Done

PR #213, opened 2018-11-07
Two fixes for the migrator code.

timuralp:
1. Fixes the call to `InternalClient.delete_object()` to not clobber acceptable statuses and properly pass headers.
2. Fixes the migrator tests to instantiate the logger exactly once.


timuralp:
Fixed up the test.

PR #214, opened 2018-11-07
decode_s3_header should return unicode strings.

timuralp:
The code that relies on decode_s3_header (e.g. the shunt module)
expects unicode, rather than byte strings.


timuralp:
I'm not sure which part is awful -- I don't think we can do anything else here, as we expect an ASCII string or a unicode string that we can pass to `unicode()` in the shunt.

PR #216, opened 2018-11-12
Fixup for the updated container_crawler library.

timuralp:
Container Crawler moved the exceptions into the utils module. The main
crawler code has been moved into a separate module as well. This allows
a called to import only exceptions, for example, without pulling in the
rest of the code.


timuralp:
Once approved, I'll update the container crawler version to a tagged release (the container crawler changes will be merged first).

PR #217, opened 2018-11-14
Implement read-only bucket validation.

timuralp:
Allows us to validate existence and access to a read-only bucket. This
is useful for migrations, where customers are unlikely to grant
PUT/DELETE/POST access to the bucket.


s3_sync/verify.py
```diff
@@ -146,9 +192,13 @@ def main(args=None):
                 swift_key = 'fabcab/cloud_sync_test'
         else:
             swift_key = 'cloud_sync_test_object'
-        result = validate_bucket(
-            provider, swift_key,
-            args.protocol == 'swift' and args.bucket == '/*')
+        if args.read_only:
+            result = validate_read_only_bucket(
+                provider, args.bucket == '/*', args.prefix)
```
thiagodasilva: in the case users specify '/*' would it be necessary to check read permissions for all buckets insteady of just one? I'm assuming users could mistakenly provide read acl just for one bucket. is that a concern?

timuralp: I don't think that makes sense, given that they could have many buckets/containers (e.g. a few 100k in the case of eBay and Swift). If/When 1space can surface errors, that would resolve IAM-style issues with bucket access. This check only ensures that the credentials are likely OK (at least for one bucket).

test/integration/test_verify.py
```diff
@@ -150,3 +154,79 @@ def test_s3_bad_bucket(self):
...
+    def test_swift_read_only(self):
+        self.assertEqual(0, s3_sync.verify.main([
+            '--protocol=swift',
+            '--endpoint=' + self.SWIFT_CREDS['authurl'],
+            '--username=' + self.SWIFT_CREDS['dst']['user'],
+            '--password=' + self.SWIFT_CREDS['dst']['key'],
+            '--read-only'
+        ]))
+
+    def test_swift_read_only_with_container(self):
```
thiagodasilva: thinking we should add a test that sets the read-only ACLs on a given bucket and test that. wdyt? is there value in that? I'd expect a test that expects failure when --read-only flag is not passed and one then expects success when --read-only is added.

timuralp: Sure. Doing that with S3Proxy is tricky, but should be doable in Swift. The test plumbing might be annoying though -- we'll need to override the storage URL and setup a container from a different account. Do you think we need to do this for this patch or can defer until after merging that patch train?

test/integration/test_verify.py
```diff
@@ -150,3 +154,79 @@ def test_s3_bad_bucket(self):
...
+        self.assertEqual(0, s3_sync.verify.main([
+            '--protocol=s3',
+            '--endpoint=' + self.S3_CREDS['endpoint'],
+            '--username=' + self.S3_CREDS['user'],
+            '--password=' + self.S3_CREDS['key'],
+            '--bucket=' + self.s3_bucket,
+            '--prefix=',
+            '--read-only']))
+
+    def test_swift_read_only(self):
```
thiagodasilva: interesting that --bucket is not required

timuralp: Good catch -- that's a buggy test. I can add a validation check that `--read-only` only makes sense if bucket is set.

PR #218, opened 2018-11-15
Implement list_buckets() for S3

timuralp:
There are a couple commits here. The goal is to implement and use the `list_buckets()` S3 API. To that end, the `BaseSync.list_buckets()` prototype was changed to avoid including the `marker`, `prefix`, `delimiter`, and `limit` parameters. These can be passed for a provider that implements them, but are not required.


s3_sync/verify.py
```diff
@@ -72,8 +72,8 @@ def validate_read_only_bucket(provider, find_bucket, prefix):
     '''
 
     if find_bucket:
-        # TODO: implement for S3
-        resp = provider.list_buckets(None, 1, None)
+        resp = provider.list_buckets(
```
thiagodasilva: how about just `list_buckets(limit=1)`

s3_sync/sync_swift.py
```diff
@@ -277,12 +276,6 @@ def list_buckets(self, marker, limit, prefix, parse_modified=True):
         for entry in resp.body:
             entry['content_location'] = self._make_content_location(
                 entry['name'])
-
```
thiagodasilva: why is this not necessary anymore

s3_sync/verify.py
```diff
@@ -179,11 +179,9 @@ def main(args=None):
 
     provider = create_provider(conf, max_conns=1)
     if not args.bucket:
```
thiagodasilva: interesting, if users doesn't pass bucket, we just verify that we can list buckets.

s3_sync/sync_swift.py
timuralp: I've found it confusing that the `parse_modified` argument ever existed. I removed its uses and standardized on always returning strings for the `last_modified` entry.

s3_sync/verify.py
timuralp: Done

s3_sync/verify.py
timuralp: Right, that was the prior behavior, so you could validate access credentials without us trying to validate bucket access.

PR #219, opened 2018-11-16
add coverage reports for integration tests

thiagodasilva:
We already have coverage for unit tests, let's add for integration tests too. Once we change our integration tests to run all in-process we will get even better coverage reports.


run_tests
```diff
@@ -40,5 +40,12 @@ docker exec swift-s3-sync timeout 40 \
              --auth-version 3 --debug stat; do sleep 0.5; done' >/dev/null 2>&1
 
 # the AWS_* environment variables allow some tests to run against real S3
-docker exec -e DOCKER=true -e AWS_IDENTITY=$AWS_IDENTITY -e AWS_SECRET=$AWS_SECRET -e AWS_BUCKET=$AWS_BUCKET  swift-s3-sync nosetests /swift-s3-sync/test/integration/
+export COVER_HTML_DIR=/swift-s3-sync/.coverhtml
+docker exec -e DOCKER=true -e AWS_IDENTITY=$AWS_IDENTITY \
+    -e AWS_SECRET=$AWS_SECRET -e AWS_BUCKET=$AWS_BUCKET \
+    swift-s3-sync /bin/bash -c "mkdir -p $COVER_HTML_DIR; \
```
mdegerne: hmm, do we need this `mkdir`? This script includes the run_unit_tests script which already creates the directory.

mdegerne: Even worse, we are over-writing the coverage report from the unit tests.

thiagodasilva: Good point about not needing to create again, but I'm ok with overwriting to keep it simple. If I run `./run_tests` I'd get integration coverage and if I run `./run_unit_tests` I'd get the unit test coverage.

thiagodasilva:
@mdegerne I'm not sure what you mean by both html reports? I might be missing something. Whenever I look at coverage reports I just bring up the index.html page and navigate to the file I'm interested in looking at the coverage for...Sometimes if I want to look at the coverage a single test is providing, I'll run that test only while passing the coverage flags to nosetests to see the coverage report.

mdegerne:
One of the things I find html coverage reports useful for, in general, is to determine if we have code that NEVER gets tested by our tests. It would be actually helpful to see the total result of unit test and integration test coverage reports so that we can determine if we are missing coverage.

thiagodasilva:
> One of the things I find html coverage reports useful for, in general, is to determine if we have code that NEVER gets tested by our tests. It would be actually helpful to see the total result of unit test and integration test coverage reports so that we can determine if we are missing coverage.

@mdegerne you actually bring up an interesting point of overlapping tests. I've been used to always trying to write both unit test and integration test for a new piece of code, but you seem to indicate that only one might be needed. Can you share more on that? I've been coming in with the assumption, so that is why I was wanting to add coverage reports for integration tests too. I'm always looking at both reports and trying to make sure we have tests on both...

mdegerne:
@thiagodasilva To my earlier point, it wasn't so much that one or the other was sufficient, but that it makes me extremely nervous to not have ANY tests cover a piece of code. Of course it is best if we have coverage from both, but we are in a state now where I am concerned that there is code not tested at all.

PR #220, opened 2018-11-19
Restore MPUs and big objects through the shunt.

timuralp:
During migrations, the shunt should be able to restore objects from S3.
We currently restore regular objects, but struggle in two circumstances:
objects that are multi-part uploads and objects bigger than the Swift
max file size limit (which is 5GB by default).

The patch introduces classes to restore each kind of objects and a
common abstraction for the large object restores (used by the SLO
restores, MPU restores, and big object restores).


timuralp:
So far I only functionally validated this patch. I'm working on tests now. Please give feedback on the approach and if you see something missing. Tests might actually fail :\

s3_sync/migrator.py
```diff
@@ -38,21 +38,20 @@
...
-                    REMOTE_ETAG)
+                    create_x_timestamp_from_hdrs, diff_container_headers,
+                    diff_account_headers, get_container_headers,
+                    get_slo_etag, get_sys_migrator_header, iter_listing,
+                    MigrationContainerStates, RemoteHTTPError,
+                    SeekableFileLikeIter, SWIFT_TIME_FMT, REMOTE_ETAG)
 from swift.common.http import HTTP_NOT_FOUND, HTTP_CONFLICT
 from swift.common import swob
 from swift.common.internal_client import UnexpectedResponse
+from swift.common import swob
```
mdegerne: Repeated import

PR #221, opened 2018-11-26
Use the max_file_size in big object migrations.

timuralp:
When migrating big objects, we should only turn them into SLOs if they
are larger than the Swift max_file_size.

In addition, there are a couple of cleanup changes to the manifest
generation code in the Migrator class and unit tests.


s3_sync/migrator.py
```diff
@@ -1007,7 +1008,8 @@ def _migrate_object(self, aws_bucket, container, key):
             self._migrate_mpu(aws_bucket, container, key, resp)
             return
         content_length = int(resp.headers['Content-Length'])
-        if content_length > self.segment_size:
+        if (content_length > swift.common.constraints
+                .EFFECTIVE_CONSTRAINTS['max_file_size']):
```
mdegerne: This can also be written:
`if (content_length > swift.common.constraints.MAX_FILE_SIZE):`

mdegerne: Possible reason not to might be to make mocking easier?

timuralp: I need to check what happens if the module has been already loaded and then the constant's value is changed. I'm fairly confident the `EFFECTIVE_CONSTRAINTS` is safe in that case. Wouldn't matter for the migrator, though, so I'll change it.

PR #222, opened 2018-11-27
Create a Swift container for shunt migration reqs.

timuralp:
We should create the Swift container in a number of requests through the
shunt. The worst case is object GET, because we fail to restore/migrate
the object. However, we should do the same during container GET/HEAD
requests.


s3_sync/shunt.py
```diff
@@ -514,6 +540,25 @@ def handle_post(
...
+        headers = dict(headers)
+        if migration:
+            headers[get_sys_migrator_header('container')] =\
+                MigrationContainerStates.MIGRATING
+        put_container_req = make_subrequest(
+            req.environ,
+            method='PUT',
+            path=container_path,
+            headers=headers,
+            swift_source='CloudSync Shunt')
```
mdegerne: Should this not be "1space shunt" as above?

timuralp: I thought about it. Eventually yes, but it's an existing line that got moved. Should I change the name? Happy to do it, but that's why I didn't change it.

mdegerne: I just feel that it is inconsistent to have two different strings for "swift_source" and that it would confuse log analysis. Either both should be "CloudSync Shunt", or both should be "1space shunt".

timuralp: Done -- fixed up the name.

PR #224, opened 2018-12-07
updated a few docs and requirements to make installation easier

notmyname:
The change in the requirements file is to exactly match Swift's
eventlet dependency


docs/README.md
```diff
@@ -57,29 +57,40 @@ Until we can merge the boto patches, you will also have to install botocore from
 our fork (do this before installing 1space):
 `pip install -e git://github.com/swiftstack/botocore.git@1.12.23.1#egg=botocore`
 
-Build the package to be installed on the nodes with:
+You also will need to install the `container-crawler` library from Git:
```
thiagodasilva: To keep things simple, I'd remove the separate notes about installing container-crawler and botocore and just keep the wording about installing both with `pip install -r requirements.txt`

docs/README.md
```diff
@@ -57,29 +57,40 @@ Until we can merge the boto patches, you will also have to install botocore from
...
-`swift-s3-sync` has to be invoked with a configuration file, specifying which
-containers to watch, where the contents should be placed, as well as a number of
-global settings. A sample configuration file is in the
+Both `swift-s3-sync` and `swift-s3-migrator` must be invoked with a
+configuration file, specifying which containers to watch, where the
+contents should be placed, as well as a number of global settings. A
+sample configuration file is in the
 [repository](https://github.com/swiftstack/1space/blob/master/sync.json-sample).
 
+Both of these tools uses the container crawler library to walk drives on
```
thiagodasilva: technically it doesn't actually walk the drives, but it does locate the container dbs using the container-ring.

PR #228, opened 2018-12-14
Enable Container Metadata sync (Swift->Swift)

mdegerne:
Requires version 0.1.1 or higher of container_crawler


s3_sync/sync_container.py
```diff
@@ -157,6 +159,21 @@ def statsd_increment(self, metric, value):
...
+    @staticmethod
+    def _hash_dict(data):
+        # Only used for metadata, so keys and values are strings
+        return md5.md5(str(sorted(data.items()))).hexdigest()
+
+    def handle_container_metadata(self, metadata_dict, db_id):
+        relevant_metadata = self.provider.select_container_metadata(
+            metadata_dict)
+        metadata_hash = self._hash_dict(relevant_metadata)
+        current_hash = self._get_status_row(self.METADATA_HASH_KEY, db_id)
```
timuralp: The default return value from `_get_status_row()` if the entry does not exist is 0. That shouldn't match any hash results, but it might better to check explicitly. I'm also not super fond of using `_get_status_row` to look up the hash value. Maybe we need to make that function more generic and catch `KeyError` in the callers. Defining `_get_metadata_hash()` now would allow us to swap some of this interface with fewer changes down the road.

s3_sync/sync_swift.py
```diff
@@ -582,6 +582,21 @@ def _check_slo_uploaded(self, key, remote_meta, internal_client,
                          % key)
         return True
 
+#    def select_container_metadata(self, metadata):
```
timuralp: Commented out code

s3_sync/sync_swift.py
```diff
@@ -582,6 +582,21 @@ def _check_slo_uploaded(self, key, remote_meta, internal_client,
...
+#        if not self.sync_container_metadata:
+#            return {}
+#        result = {k: metadata[k] for k in metadata
+#                  if k.startswith('x-container-meta')}
+#        return result
+#
+    def update_container_metadata(self, metadata):
```
timuralp: For consistency with the APIs, let's use `post_container` (or `post_bucket`).

s3_sync/sync_swift.py
```diff
@@ -582,6 +582,21 @@ def _check_slo_uploaded(self, key, remote_meta, internal_client,
...
+        with self.client_pool.get_client() as swift_client:
```
timuralp: Let's standardize on using `_call_swiftclient()`.

s3_sync/sync_s3.py
```diff
@@ -853,6 +853,14 @@ def _get_error_message(self, error, op, args):
             parts.append(error.message)
         return ' '.join(parts)
 
+    @staticmethod
+    def select_container_metadata(metadata_dict):
```
timuralp: The `BaseSync` method is not a static method. Probably should pass `self` for consistency?

s3_sync/sync_s3.py
```diff
@@ -853,6 +853,14 @@ def _get_error_message(self, error, op, args):
...
+        return {}
+
+    @staticmethod
+    def update_container_metadata(metadata):
```
timuralp: Also, shouldn't be a `static_method` and see comments on adding `post_container` in `SyncSwift`.

s3_sync/sync_swift.py
```diff
@@ -582,6 +582,21 @@ def _check_slo_uploaded(self, key, remote_meta, internal_client,
...
+            try:
+                swift_client.post_container(
+                    self.remote_container, headers=metadata)
+            except Exception as e:
+                self.logger.info("Container Post failed: %s", e)
```
timuralp: The caller should be able to check whether this succeeded or failed (even if today's callers don't care).

s3_sync/base_sync.py
```diff
@@ -242,6 +245,18 @@ def __repr__(self):
             self.aws_bucket,
         )
 
+    def select_container_metadata(self, metadata):
```
thiagodasilva: should add unit test for `select_container_metadata`

s3_sync/sync_container.py
thiagodasilva: just a nit but could we rename `current` to `previous` or `old` or `stored`. it would make the distinction to `metadata_hash` stand out more

s3_sync/base_sync.py
```diff
@@ -242,6 +245,18 @@ def __repr__(self):
...
+        result = dict()
+        if not self.sync_container_metadata:
+            return result
+        for key in self.RELEVANT_CONTAINER_METADATA:
+            if key in metadata:
+                result[key] = metadata[key][0]
```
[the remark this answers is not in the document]
mdegerne: Adding comments

s3_sync/sync_container.py
```diff
@@ -157,6 +159,21 @@ def statsd_increment(self, metric, value):
         self.statsd_client.update_stats(
             '.'.join([self.statsd_prefix, metric]), value)
 
+    @staticmethod
```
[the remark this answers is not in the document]
mdegerne: ok

s3_sync/sync_container.py
mdegerne: implementing _get_metadata_hash(db_id), but not changing _get_status_row at this time.

s3_sync/sync_swift.py
mdegerne: removing

s3_sync/sync_swift.py
mdegerne: ok

s3_sync/sync_swift.py
mdegerne: ok

s3_sync/sync_s3.py
mdegerne: removing staticmethod headers all around

s3_sync/sync_swift.py
mdegerne: improving this

s3_sync/base_sync.py
mdegerne: working on this.

s3_sync/sync_container.py
mdegerne: changing to `last_hash`

s3_sync/base_sync.py
```diff
@@ -246,6 +246,16 @@ def __repr__(self):
...
 
     def select_container_metadata(self, metadata):
+        """Select container metadata to track
+
+        This is used to select which metadata to care about with regards to
+        sync'ing with the remote cluster. It also translates from the database
+        representation to a header format.
+        :param metadata: in the format from broker.metadata, that is a dict of
+                         the form {<key>: (<value>, <timestamp>), ...)
+        :returns: dict of the form {<key>: <vlaue>, ...) containing only
```
thiagodasilva: nit: value
if no other changes are needed, can be fixed in a follow-on commit

s3_sync/base_sync.py
```diff
@@ -242,6 +245,28 @@ def __repr__(self):
...
+                  relevant key value pairs
+        """
+        result = dict()
```
thiagodasilva: again just another nit...but curious about the preference for `result = dict()` over `result = {}`. According to [1] `dict()` can be 6 times slower than {}. The other argument would be to default to a standard in the code...

[1] - https://doughellmann.com/blog/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/

mdegerne:
No strong reason. I think I have done it this way elsewhere. I'll fix up
the nits.

test/unit/test_base_sync.py
```diff
@@ -208,6 +208,28 @@ def test_double_release(self, factory_mock):
             client.close()
         self.assertEqual(1, client.semaphore.balance)
 
+    @mock.patch('s3_sync.base_sync.BaseSync._get_client_factory')
+    def test_select_container_metadata(self, factory_mock):
+        factory_mock.return_value = mock.Mock()
+        test_headers = {
+            'X-Container-Meta-foo': ('foo', '1545179217.201453'),
+            'x-delete-at': ('1645179217', '1545179217.201453'),
+            'someotherkey': ('boo', '1545179217.201453'),
```
thiagodasilva: maybe use x-versions-location

s3_sync/sync_container.py
```diff
@@ -157,6 +164,30 @@ def statsd_increment(self, metric, value):
...
+
+    def handle_container_metadata(self, metadata_dict, db_id):
+        relevant_metadata = self.provider.select_container_metadata(
+            metadata_dict)
+        metadata_hash = hash_dict(relevant_metadata)
+        last_hash = self._get_metadata_hash(db_id)
+        if last_hash == metadata_hash:
+            return
+        post_resp = self.provider.post_container(relevant_metadata)
+        # Not all providers return a ProviderResponse (some return None)
```
timuralp: @mdegerne can you turn this into a TODO and specify the problematic provider? The interface should be unified.

s3_sync/sync_s3.py
```diff
@@ -853,6 +853,12 @@ def _get_error_message(self, error, op, args):
             parts.append(error.message)
         return ' '.join(parts)
 
+    def select_container_metadata(self, metadata_dict):
+        return {}
+
+    def post_container(self, metadata):
+        pass
```
timuralp: I'm guessing this is the provider that returns None. Can we return a canned failure here, like:
```suggestion
         return ProviderResponse(False, 501, {}, '')
```

s3_sync/sync_container.py
```diff
@@ -157,6 +164,30 @@ def statsd_increment(self, metric, value):
         self.statsd_client.update_stats(
             '.'.join([self.statsd_prefix, metric]), value)
 
+    def _get_metdata_hash(self, db_id):
```
thiagodasilva: typo: _get_metadata_hash

mdegerne:
Thanks.

test/integration/utils.py
```diff
@@ -108,6 +108,9 @@ def _save_status_row(self, row, field_id, db_id):
             self._status[db_id] = status
         status[field_id] = row
 
+    def _get_metadata_hash(self, db_id):
```
thiagodasilva: i don't think this method is needed...

PR #229, opened 2018-12-17
Propagate listing errors through the shunt.

timuralp:
The shunt should return the listing error from the remote provider, as
opposed to quietly swallowing it, leading to unexpected results.

This means that both the Swift list request and the remote provider must
be successful in order to list linked containers.


timuralp:
This bumps the coverage for the shunt in unit tests above 70%! Small victory :)

timuralp:
@mdegerne discussed this internally and going to defer this patch until we allow for an escape hatch to still be able to list objects even if the remote blobstore is experiencing failures.

PR #230, opened 2018-12-20
Add support for syncing sharded containers

thiagodasilva:
Add integrated tests to cover syncing sharded containers. Tests cover two scenarios:
1. Start syncing after container has already been sharded
2. Start syncing before container has been sharded and continue after sharding


containers/swift-s3-sync/launch.sh
```diff
@@ -35,6 +39,7 @@ mkdir -p /srv/1/node/sdb1 /srv/2/node/sdb2 /srv/3/node/sdb3 /srv/4/node/sdb4 \
 cd /swift-s3-sync; pip install -e .
 
 /usr/bin/sudo -u swift /swift/bin/startmain
+# TODO: /usr/bin/sudo -u swift /swift/bin/startrest
```
thiagodasilva: remove this

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,138 @@
...
+        self.assert_container_delete_fails(conn, self.cont)
+
+        # shrink 1st half
+        self.sharders.once()
+
+        # shrink 2nd half
+        self.sharders.once()
+        self.assert_container_object_count(conn, self.cont, 0)
+
+        # last container delete is run by tearDownClass in __init__
```
thiagodasilva: fix this, since we are creating a different container here (see Line 61), we should delete it here too

test/integration/utils.py
```diff
@@ -138,7 +138,7 @@ def instance(self, settings, per_account=True):
 
 def get_container_crawler(profile, **kwargs):
     defaults = {
-        'log_level': 'info',
+        'log_level': 'debug',
```
thiagodasilva: need to change this back

test/integration/test_migrator.py
```diff
@@ -1481,6 +1481,13 @@ def _make_manifest(index):
     def test_migrate_many_dlos(self):
         '''Ensure we can migrate many DLOs'''
         migration = self.swift_migration()
+
+        # the root_ prefix is added automatically by BaseSync
+        # Adding it here since we will use this config later
+        # for making assertions and the config must match
+        # with the copy in the migrator
+        migration['root_account'] = migration['account']
```
thiagodasilva: forgot to remove this

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+        ret = find_replace_shard_ranges(broker, args)
+        if ret != 0:
+            self.fail('failed to create shard ranges')
+
+        # 1space will not synchronize objects when sharding
+        # has started, but objects are still in root container
+        # and not yet in its own shard db. The objects that
+        # already made to its shard are synced.
+        self.sharders.once()  # create 1st and 2nd shard
+        self.sharders.once()  # create 3rd shard
```
timuralp: Can we run the test with fewer shards? Does it matter?

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+        self.assert_container_object_count(conn, self.cont, object_count)
+
+    def tearDown(self):
+        conn = self.conn_for_acct(self.acct)
+        clear_swift_container(self.swift_src, self.cont)
+        self.assert_container_delete_fails(conn, self.cont)
+
+        # after deleting objects, run sharder to shrink back to root
+        # then delete
+        self.sharders.once()
```
timuralp: Can you add a note about why the sharder needs to be run twice?

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+        '''
+        Test syncing *for the first time* after container
+        has already been sharded
+        '''
+        target_conn = self.conn_for_acct_noshunt(self.mapping['aws_account'])
+
+        # shard container
+        part, nodes = CONTAINER_RING.get_nodes(self.acct,
+                                               self.cont)
+        broker = self._get_broker(
```
timuralp: I'd sink the broker stuff into the `_shard_container()`. There is no other reason for us to get the DB broker, right?

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+    def setUp(self):
+        self.mapping = self._find_mapping(
+            lambda m: m['aws_bucket'] == 'sharded-target')
+
+        self.acct = self.mapping['account']
+        self.cont = '%s_%s' % (self.mapping['container'], uuid4().hex)
+        self.mapping['container'] = self.cont
+
+        conn = self.conn_for_acct(self.acct)
+        object_count = 60
```
timuralp: I'd make this a top level constant. Also, do you know how long the test would take with fewer object, say 30? Maybe there is benefit to reducing this?

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+        # and not yet in its own shard db. The objects that
+        # already made to its shard are synced.
+        self.sharders.once()  # create 1st and 2nd shard
+        self.sharders.once()  # create 3rd shard
+        self.updaters.once()  # create .shards_ account db
+
+    def _upload_objects(self, conn, object_count):
+        for i in range(object_count):
+            obj_uuid = uuid4().hex
+            obj_name = 'obj_%s' % obj_uuid
```
timuralp: Using the index (`i`) might be useful if we ever want to try to look at the container and debug it.

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+            headers = target_conn.head_object(self.mapping['aws_bucket'], obj)
+            self.assertEqual(etag, headers['etag'])
+
+        # test that statsd server got the correct results for root container
+        self._assert_stats(self.mapping, len(self.test_args), 'copied_objects')
+        clear_swift_container(self.swift_dst, self.mapping['aws_bucket'])
+
+    def test_shard_after_initial_sync(self):
+        '''
+        Test syncing before sharding, then sync again after sharding has
```
timuralp: Is there a code path that this test covers that the test above doesn't?

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # sanity test
+        self.sharders.once()  # update container count
+        self.assert_container_object_count(conn, self.cont, 90)
+
+        # sync to target again
+        crawler.run_once()
+        for obj, body, etag in self.test_args:
+            headers = target_conn.head_object(self.mapping['aws_bucket'], obj)
+            self.assertEqual(etag, headers['etag'])
```
timuralp: If we're only comparing ETags, consider doing a container listing.

timuralp: Do we not need to run `clear_swift_container` here too? What about checking stats?

containers/swift-s3-sync/swift-s3-sync.conf
```diff
@@ -218,6 +217,19 @@
             "protocol": "swift",
             "retain_local": true
         },
+        {
```
timuralp: What if we don't add this here, but rather specify the mapping explicitly in the test? I'd like to start moving to that model anyway, as it's a bit wonky to maintain this file for every possible test combination.

test/integration/test_sync_sharded_container.py
```diff
@@ -0,0 +1,191 @@
...
+        # shard container
+        part, nodes = CONTAINER_RING.get_nodes(self.acct,
+                                               self.cont)
+        broker = self._get_broker(
+            part, nodes[0], self.acct, self.cont)
+        self._shard_container(broker, 60)
+
+        # add 30 more objects
+        conn = self.conn_for_acct(self.acct)
+        self._upload_objects(conn, 30)
```
timuralp: I wonder if we could use fewer objects. Do you know the per-request cost (since we're doing these serially)?

test/unit/test_verify.py
```diff
@@ -27,6 +27,7 @@
 
 @mock.patch('s3_sync.verify.validate_bucket', return_value=object())
 class TestMainTrackProvider(unittest.TestCase):
+
```
timuralp: Spurious white space change?

test/integration/utils.py
timuralp: We can supply this per-test, so you shouldn't have to change the defaults.

thiagodasilva: done

containers/swift-s3-sync/launch.sh
```diff
@@ -12,6 +12,10 @@ find /var/run/ -name *.pid -delete
 # proxy-server instance on port 8082.
 cp -f /swift-s3-sync/containers/swift-s3-sync/{internal-client,proxy-server,proxy-server-noshunt}.conf /etc/swift/
 
+# TODO: add this to bouncestorage/swift-aio
+printf '\n[container-sharder]\n' >> /etc/swift/container-server/1.conf
```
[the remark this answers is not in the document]
thiagodasilva: for such a small change and it's just the default value, I'd suggest making the change in swift-aio[1] instead of creating another file here for us to manage...OTOH we could set other (non-default) options if we really wanted to... :thinking: 

[1] - https://github.com/bouncestorage/docker-swift/blob/master/swift/container-server/1.conf

test/integration/__init__.py
```diff
@@ -22,13 +22,16 @@
...
 from swift.common.middleware.acl import format_acl
+from swift.common.ring import Ring
 import swiftclient
 import time
 import unittest
 import urllib
 
 import utils
 
+CONTAINER_RING = Ring('/etc/swift', ring_name='container')
```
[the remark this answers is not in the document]
thiagodasilva: will fix.

PR #238, opened 2019-01-09
Rename the container and executables to 1space-*.

timuralp:
Renamed the entry points and main container to 1space (from swift-s3-*).


timuralp:
We need to also update docs, but I didn't want to collide with the docs PR, so I left them alone.

setup.py
```diff
@@ -17,9 +17,9 @@
       install_requires=['boto3==1.9.23'],
       entry_points={
           'console_scripts': [
-              'swift-s3-sync = s3_sync.__main__:main',
-              'swift-s3-verify = s3_sync.verify:main',
-              'swift-s3-migrator = s3_sync.migrator:main',
+              '1space-sync = s3_sync.__main__:main',
+              '1space-verify = s3_sync.verify:main',
+              '1space-migrator = s3_sync.migrator:main',
```
tipabu: Should we also rename the package name up on line 5? I worry about breaking already-deployed configs on upgrade, though...

timuralp: We should. I was trying to stage it in smaller steps. Renaming daemons would be a relatively small change for existing deployments, but renaming the package affects the proxy server configuration :\

PR #239, opened 2019-01-11
Add a callback to audit bytes transferred.

timuralp:
The patch allows 1space to keep track of bytes transferred during an
upload and report it via statsd.


s3_sync/sync_container.py
```diff
@@ -199,7 +199,10 @@ def handle(self, row, swift_client):
             _, _, meta_ts = decode_timestamps(row['created_at'])
             if time.time() <= self.copy_after + meta_ts.timestamp:
                 raise RetryError('Object is not yet eligible for archive')
-            status = self.provider.upload_object(row, swift_client)
+            status = self.provider.upload_object(
```
thiagodasilva: couple issues with this approach, i think #1 being worse than #2:
1. it's counting bytes regardless of object upload being successful or not. If upload fails 1/2 way it will still count the bytes that were sent and I'm not sure we want that. Plus it would now create a discrepancy with 'copied_objects' which is only updated on a successful upload.
2. plumbing the callback all the way down to the FileWrapper feels a bit hacky and again very different from how `copied_objects` is being incremented. I was hoping upload_object could return more than the status and maybe return an object from where we could get the status and the content-length (i.e., similar to Response in swift)

timuralp: I think this gets back to why we want to do this: 1space should have a way to understand its bandwidth usage. If we emit total bytes transferred only on success and only on completion, there could be both large windows where no data is available (e.g. large object uploads) and missing data (an upload that results in an error still uses the network).

Happy to rename the value of the stat, but stand by this approach. I don't think emitting stats only on completion will actually satisfy the ask and answer the question of "how much bandwidth is 1pace using right now?".

mdegerne: For my $0.02, I think that both stats are important (bytes transferred and bytes successfully copied). I think the difference in the two values would be instructive about what is working efficiently and not, where as the bytes copied is useful to have a gauge of how much data on the target actually came from the source. In all cases, good documentation of what the stats actually mean is vital.

timuralp: Sure, agreed with the documentation part. This patch implements the in-progress bytes transferred though. We can add another stat for the completed uploads. I do wonder if we ever build the dashboard view of data in the remote blob store whether it would still be useful. Maybe it would still be hard to tell where the objects originated, which is fair.

thiagodasilva: Ah! fair enough. I made the wrong assumption that we were shooting for the bytes successfully copied, thanks for clarifying. My guess is that (as @mdegerne already mentioned) that may also be useful at some point but we can get to it in a follow up patch.

PR #244, opened 2019-01-18
Set remote_delete_after on the S3 bucket prefix based on configuration

mdegerne:
This is equivalent functionality to existing swift remote_delete_after setting.

Note: Added a "RUN_LONG_TESTS" env variable check for multipart migration test.


s3_sync/sync_s3.py
```diff
@@ -39,6 +40,31 @@
     SeekableFileLikeIter)
 
 
+def prefix_from_rule(rule):
+    # Note: Prefix at Rule level is depreciated, so Filter has
```
timuralp: Nit: deprecated (vs depreciated)

s3_sync/sync_s3.py
```diff
@@ -39,6 +40,31 @@
...
+    # precidence
```
timuralp: Nit: precedence (vs precidence)

s3_sync/sync_s3.py
```diff
@@ -143,6 +169,82 @@ def _close_conn(conn):
             if close:
                 close()
 
+    def _get_lifecycle_configuration_for_bucket(self):
+        # get existing lifecycleconfiguration
+        # if doesn't exist, need to write a whole bucket policy otherwise
```
timuralp: This is something we should track down.  I'd at least turn it into a TODO.

s3_sync/sync_s3.py
```diff
@@ -143,6 +169,82 @@ def _close_conn(conn):
...
+
+    def _update_expire_after(self):
+        remote_del_after = self.settings.get('remote_delete_after', 0)
+        if not remote_del_after:
+            return ProviderResponse(True, 200, {}, '')
+        rules = self._get_lifecycle_configuration_for_bucket()
+        prefix = self.get_s3_name('')
+        days = int(remote_del_after / (60 * 60 * 24.0) + 0.5)
+
+        conflicts = filter(partial(is_conflict, prefix), rules)
```
timuralp: We could also do this with lambda `filter(lambda rule: is_conflict(prefix, rule), rules)`. I don't think it makes a big difference one way or another.

s3_sync/sync_s3.py
```diff
@@ -143,6 +169,82 @@ def _close_conn(conn):
...
+            rules = s3_status['Rules']
+        return rules
+
+    def _update_expire_after(self):
+        remote_del_after = self.settings.get('remote_delete_after', 0)
+        if not remote_del_after:
+            return ProviderResponse(True, 200, {}, '')
+        rules = self._get_lifecycle_configuration_for_bucket()
+        prefix = self.get_s3_name('')
+        days = int(remote_del_after / (60 * 60 * 24.0) + 0.5)
```
timuralp: Consider using a constant for 1 day or even the `timedelta` module (`datetime.timedelta(days=1).total_seconds()`).

s3_sync/sync_s3.py
```diff
@@ -143,6 +169,82 @@ def _close_conn(conn):
...
+        else:
+            rules.append(newrule)
+        self._put_lifecycle_configuration(rules)
+        message = 'Created new lifecycle rule for bucket (%s): %s'\
+            % (self.aws_bucket, newrule)
+        self.logger.info(message)
+        return ProviderResponse(True, 200, {}, message)
+
+    def _put_lifecycle_configuration(self, rules):
+        with self.client_pool.get_client() as s3_client:
```
timuralp: Can we use `_call_boto()`?

s3_sync/sync_s3.py
```diff
@@ -853,7 +955,14 @@ def _get_error_message(self, error, op, args):
             parts.append(error.message)
         return ' '.join(parts)
 
+    def post_container(self, metadata):
+        # NOTE: If metadata used, remove 'X-SS-REMOTE-DELETE-AFTER' key
```
timuralp: Can you expand on the NOTE a little bit? I'm not sure I'm following the comment about removing the key.

I see why we did it this way (so we can hook into the existing `handle_container_metadata` method in `SyncContainer`), but wonder if this is problematic, because in general there is no `POST` on an S3 bucket and this method now suggests some support for it. Maybe we could have special cased S3 in `SyncContainer.handle_container_metadata()` and made the `update_lifecycle_policy` method public?

s3_sync/sync_s3.py
```diff
@@ -143,6 +169,82 @@ def _close_conn(conn):
...
+                'Filter': {'Prefix': ''},
+                'ID': 'fakerule',
+                'Status': 'Disabled'}
+            self._put_lifecycle_configuration([fake_rule])
+            rules = []
+        else:
+            rules = s3_status['Rules']
+        return rules
+
+    def _update_expire_after(self):
```
timuralp: I think `_update_lifecycle_policy` would be more explicit. I suspect, in the future, we may have asks to add other lifecycle configuration (e.g. transition to glacier) and that would make the method more explicit about what it's doing.

test/integration/test_s3_sync.py
```diff
@@ -1038,3 +1039,58 @@ def ensure_remote():
...
+        match = next((r.get('Filter', {}).get('Prefix') ==
+                      prefix for r in orig_rules), None)
+        if match:
+            old_del_after = match.get('Expiration', {}).get('Days')
+            if old_del_after:
+                mapping['remote_delete_after'] = \
+                    (old_del_after + 1) * 60 * 60 * 24
+
+        del_after = mapping['remote_delete_after']
+        exp_days = int(del_after / (60 * 60 * 24.0) + 0.5)
```
timuralp: I'd use a constant for 1 day here again. We could also use `round()`, as opposed to `+ 0.5`. Don't have a preference on the use of `round()` though.

test/integration/test_s3_sync.py
```diff
@@ -1038,3 +1039,58 @@ def ensure_remote():
...
+        crawler = utils.get_container_crawler(mapping)
+        crawler.run_once()
+
+        rules = provider._get_lifecycle_configuration_for_bucket()
+
+        # clean up before checking if it worked (in case it didn't)
+        conn.delete_container(mapping['container'])
```
timuralp: This is more of technical debt stuff that I'm hoping we can address in the future: move all clean up like this into `tearDown` and have a way of `checkingOut` a container. That would make handling failures much cleaner.

test/integration/test_s3_sync.py
```diff
@@ -1038,3 +1039,58 @@ def ensure_remote():
...
+        orig_rules = provider._get_lifecycle_configuration_for_bucket()
+        prefix = provider.get_s3_name('')
+
+        match = next((r.get('Filter', {}).get('Prefix') ==
+                      prefix for r in orig_rules), None)
+        if match:
+            old_del_after = match.get('Expiration', {}).get('Days')
+            if old_del_after:
+                mapping['remote_delete_after'] = \
+                    (old_del_after + 1) * 60 * 60 * 24
```
timuralp: I'd use a constant for 1 day.

s3_sync/sync_s3.py
mdegerne: fixed

s3_sync/sync_s3.py
mdegerne: fixed

s3_sync/sync_s3.py
mdegerne: Added a TODO in the code.

s3_sync/sync_s3.py
mdegerne: good call. Fixed

s3_sync/sync_s3.py
mdegerne: fixing

s3_sync/sync_s3.py
mdegerne: Fair enough. It's good to think about both ways. I'm going to leave as is for now.

test/integration/test_s3_sync.py
mdegerne: Round would leave it as float and I would still need to convert to int. Adding constants, though.

test/integration/test_s3_sync.py
mdegerne: Agree. For things outside of "standard" set of mappings (like this), we may still need some special processing, but it can probably be managed.

s3_sync/sync_s3.py
```diff
@@ -39,6 +40,31 @@
...
+        r_prefix = r_filter.get('Prefix')
+    if r_prefix is None:
+        r_prefix = rule.get('Prefix')
+    return r_prefix
+
+
+def is_conflict(prefix, rule):
+    r_prefix = prefix_from_rule(rule)
+    if r_prefix is not None:
+        if not prefix.startswith(r_prefix):
```
thiagodasilva: wondering if this is an issue:
1space might create a rule like this:
`prefix 32911e/AUTH_test/aws-source-test1/`
and user might create a rule with a slightly different prefix:
`prefix 32911e/AUTH_test`

Current code would not catch this as a conflict, but i think it should be...thoughts?

mdegerne: How would that not be caught? In this case prefix starts with the rule prefix, which is precisely what we are checking for. That said, I didn't try to be completely exhaustive as the S3 checking will likely be more complete already and we log the error returned from S3.

thiagodasilva: I think the issue is that I manually created the "conflicting" rule after the one created by 1space, so if they are returned in order of creation, I think line 211 is returning as it encounters the 1space rule first. In fact, I even tried modifying my own rule to "Whole bucket" and 1space does not catch it as a conflict.

mdegerne: Ah. That's a different issue than I was thinking of. It's interesting that S3 now permits you to create the conflicting rules at all (given that all the research I did indicated that you couldn't). That was the assumption built in to the code to begin with. Given that, I may just change this section to merely log a list of conflicting rules and attempt to set the value anyway. If S3 doesn't like it, the S3 message will be logged as well as all the conflicts, which should make deciding what to do simpler for Support.

s3_sync/sync_s3.py
mdegerne: fixed.

s3_sync/sync_s3.py
mdegerne: I think this is fundamentally a naming issue. I don't want to put provider specific logic in sync_container, because that defeats the value of using providers. Initial thought is to change function name to `update_container_metadata` or maybe `container_metadata_changed`. That keeps the intent of recognizing a change in the relevant metadata, but removes the explicit definition of what the provider does with it.

timuralp: We have to keep the high level `post_container` (in `BaseSync`), but not every provider supports it. In this case, S3 doesn't have a notion of POST on a bucket and passing arbitrary metadata to this method doesn't do what the caller intends. We're also abusing the `handle_container_metadata` here already -- the _metadata_ is pulled out of settings and not actually related to the Swift container. I think when the abstraction doesn't map onto the blobstore, we have to poke a hole (hence the suggestion to special case S3). I'm ok cleaning it up afterward, though.

Thanks for updating the note!

PR #246, opened 2019-01-24
Fix logging for post_container failures.

timuralp:
The exc_info is a tuple of type, value, traceback that we have to handle
correctly (otherwise the message is not informative).


s3_sync/sync_container.py
```diff
@@ -186,8 +187,10 @@ def handle_container_metadata(self, metadata_dict, db_id):
             self._save_status_row(metadata_hash, self.METADATA_HASH_KEY, db_id)
             self.logger.debug(
                 'Container metadata updated for %s' % (self.aws_bucket,))
-        elif post_resp.exc_info:
-            self.logger.info(post_resp.exc_info)
+        else:
+            self.logger.error('Failed to update container metadata -- %s: %s\n%s' % (
+                post_resp.wsgi_status, post_resp.exc_info[1],
+                ''.join(traceback.format_tb(post_resp.exc_info[2]))))
```
thiagodasilva: I've been wondering if we should log the actual traceback at the debug level. Here we would have two log statements, one error with the summarized information and error message, hopefully enough for operators to understand and take action and then a second debug log with the traceback.

timuralp: Sure. I guess we could go through a fix up formatting in this way for all of the cases like this.

PR #247, opened 2019-01-26
Ensure POSTs requests are only shunted on migration profile

thiagodasilva:
POST requests should only be shunted for migration profiles and
only when the local container/object response returns a 404. Modified
test code to ensure we test correct behavior.


timuralp:
Closed via 0e7b51dce345e3ff77099eaa7f08dba243046cfb.

PR #248, opened 2019-01-26
Clean up copyright blocks

thiagodasilva:
Copyright blocks were showing up in the published docs


thiagodasilva:
shunt.py was intentionally skipped as to not cause conflict with another PR

PR #249, opened 2019-01-26
Set Content-Length on errors from S3.

timuralp:
When we are returning error messages from S3, we have to update the
Content-Length header.


mdegerne:
_update_lifecycle_policy doesn't build a WSGI/HTTP response. I'm not clear on the question, @thiagodasilva.

thiagodasilva:
cleared up my confusion with @mdegerne

PR #255, opened 2019-02-05
Optionally convert DLO to SLO if Swift->Swift.

timuralp:
When propagating objects to a Swift cluster, we should have an option to
convert Dynamic Large Objects into Static Large Objects. By default,
this option is disabled, however, it gives an operator the ability to
move away from using DLOs if desired.


s3_sync/sync_swift.py
```diff
@@ -592,24 +603,37 @@ def _keys_generator():
...
-                                         self.container,
-                                         key, swift_req_hdrs)
-            put_headers = self._get_user_headers(wrapper_stream.get_headers())
+        if self.convert_dlo:
+            manifest_body = json.dumps(manifest)
+            etag = None
+            put_headers = self._get_user_headers(internal_meta)
+            query_string = 'multipart-manifest=put'
+        else:
+            # The DLO manifest might be non-zero sized, so we retrieve the
```
thiagodasilva: In the case of converting to SLO, the dlo manifest content (if it's non-zero) might be lost, we should document that.

s3_sync/sync_swift.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, *args, **kwargs):
         if not self.remote_delete_after and self.settings.get(
                 'propagate_expiration'):
             self.propagated_headers.add('x-delete-at')
+        self.convert_dlo = self.settings.get('convert_dlo', False)
```
thiagodasilva: add documentation for new option

s3_sync/sync_swift.py
```diff
@@ -592,24 +603,37 @@ def _keys_generator():
...
+            put_headers = self._get_user_headers(internal_meta)
+            query_string = 'multipart-manifest=put'
+        else:
+            # The DLO manifest might be non-zero sized, so we retrieve the
+            # original Swift object.
+            manifest_body = FileWrapper(
+                internal_client, self.account, self.container, key,
+                swift_req_hdrs)
+            etag = manifest_body.get_headers()['etag']
+            put_headers = self._get_user_headers(manifest_body.get_headers())
```
thiagodasilva: ok, we use manifest_body.get_headers() instead of internal_meta here...

test/integration/test_s3_sync.py
```diff
@@ -1138,10 +1138,146 @@ def test_swift_dlo_metadata_update(self):
...
+
+        self._assert_stats(mapping, {'copied_objects': 1,
+                                     'bytes': sum(map(len, content))})
+        self.local_swift('post_object', mapping['container'], key,
+                         {u'x-object-meta-new-valu\xe3': u'valu\xe3',
+                          u'x-object-manifest': manifest})
+
+        # Re-processing the object should be a POST (which does not count as
+        # copied).
+        crawler = utils.get_container_crawler(mapping)
```
thiagodasilva: this can be removed

test/integration/test_s3_sync.py
```diff
@@ -1138,10 +1138,146 @@ def test_swift_dlo_metadata_update(self):
...
+        crawler.run_once()
+
+        headers = self.remote_swift('head_object', mapping['aws_bucket'], key)
+        self.assertEqual(u'valu\xe3',
```
thiagodasilva: that's pretty awesome

test/integration/test_s3_sync.py
```diff
@@ -1138,10 +1138,146 @@ def test_swift_dlo_metadata_update(self):
...
+        self.assertEqual('"%s"' % dlo_etag, headers['etag'])
+        self.assertEqual(dlo_etag,
+                         headers.get('x-object-meta-swift-source-dlo-etag'))
+        self.assertTrue(headers.get('x-static-large-object'))
+
+        self._assert_stats(mapping, {'copied_objects': 1,
+                                     'bytes': sum(map(len, content))})
+
+        # Re-processing the object should be a NOOP
+        crawler = utils.get_container_crawler(mapping)
```
thiagodasilva: can be removed...

test/unit/test_sync_swift.py
```diff
@@ -1240,9 +1221,304 @@ def fake_swift_head(container, key, **kwargs):
...
+            [mock.call(
+                self.aws_bucket, dlo_key,
+                headers={
+                    'x-object-meta-swift-source-dlo-etag': dlo_etag,
+                    'x-object-meta-new-key': 'value',
+                })],
+            mock_swift.return_value.post_object.mock_calls)
+
+    @mock.patch('s3_sync.sync_swift.swiftclient.client.Connection')
+    def test_upload_changed_dlo_to_slo(self, mock_swift):
```
thiagodasilva: Please add some docstrings with some description of what this is testing, it's really difficult to follow and figure out exactly what's being tested. With some hints it would make it easier to understand.

s3_sync/sync_swift.py
timuralp: Done -- added to the docs.

s3_sync/sync_swift.py
timuralp: Done

test/integration/test_s3_sync.py
timuralp: Done

test/integration/test_s3_sync.py
timuralp: We have to keep this for the object to be re-processed (there is no other good way to reset the status, so we re-create the crawler instance).

test/unit/test_sync_swift.py
timuralp: Done

PR #256, opened 2019-02-06
Sync Container ACL

thiagodasilva:
Added new option (`sync_container_acl`) to sync container ACLs.
Added as a separate option but still dependant on
`sync_container_metadata` also being set to ``True``. The default for
both options is ``False``.


thiagodasilva:
@mdegerne notice I changed the constant ``RELEVANT_CONTAINER_METADATA`` to a property, let me know what you think.

PR #257, opened 2019-02-07
Fixup for the common interface.

timuralp:
Fixes the common interface for put_object and adds the optional bucket
argument for the delete_object call.

Also, for the SyncSwift, moves delete_object() to calling
_call_swiftclient, so we can standardize on response and exception
handling.


s3_sync/base_sync.py
```diff
@@ -280,31 +280,31 @@ def post_object(self, swift_key, headers):
...
-        The headers provided are assumed to be in Swift parlance, and will be
-        converted to S3-style headers if necessary.
-
-        The `body` argument can be a string or unicode instance, a file-like
-        object (like a wsgi.input stream wrapped by an InputProxy instance), or
-        an iterable.
-
-        The optional query_string is used only for the Swift provider and is
-        sent on, verbatim, to the underlying swiftclient put_object() call.
+        :swift_key: object key.
```
thiagodasilva: s/swift_key/param key

s3_sync/base_sync.py
```diff
@@ -280,31 +280,31 @@ def post_object(self, swift_key, headers):
...
     def upload_object(self, row, internal_client):
         raise NotImplementedError()
 
     def update_metadata(self, swift_key, swift_meta, remote_metadata={},
-                        container={}):
+                        bucket=None):
         raise NotImplementedError()
 
-    def delete_object(self, swift_key):
+    def delete_object(self, swift_key, bucket=None):
```
thiagodasilva: and here

s3_sync/base_sync.py
```diff
@@ -280,31 +280,31 @@ def post_object(self, swift_key, headers):
...
+        :returns: an instance of ProviderResponse.
         """
         raise NotImplementedError()
 
     def upload_object(self, row, internal_client):
         raise NotImplementedError()
 
     def update_metadata(self, swift_key, swift_meta, remote_metadata={},
-                        container={}):
+                        bucket=None):
```
thiagodasilva: should we also change from swift_key and swift_meta to key and metadata here?

s3_sync/sync_swift.py
```diff
@@ -162,25 +165,23 @@ def upload_object(self, row, internal_client):
...
+            bucket = self.remote_container
+        resp = self.head_object(swift_key, bucket=bucket)
+        if not resp.success:
+            if resp.status == 404:
+                return resp
+            resp.reraise()
 
         delete_kwargs = {'headers': self._client_headers()}
-        if check_slo(headers):
+        if check_slo(resp.headers):
```
thiagodasilva: just realized we don't handle dlo here, is that a TODO item?

s3_sync/base_sync.py
timuralp: Done

s3_sync/base_sync.py
timuralp: Sure, done.

s3_sync/base_sync.py
timuralp: Done

s3_sync/sync_swift.py
timuralp: Good point! Let's do that in a separate patch. I'll make a PT story.

timuralp:
I rebased and fixed up all of the instances of `swift_key` vs `key`.

PR #259, opened 2019-02-08
Implement support for minimum segment size.

timuralp:
Allows for conversion of objects with many small segments into ones composed of larger segments. This applies both to DLO and SLO. In case of DLO, we do not allow minimum segment size to be set without also setting the convert_dlo option (converting DLOs to SLOs). This is because we cannot match the DLO prefix and naming scheme otherwise.


s3_sync/sync_swift.py
```diff
@@ -445,13 +450,88 @@ def _upload_object(self, src_container, dst_container, row,
...
+                'Combined upload of %s returned mismatching ETags; '
+                'removing the destination object %s/%s' % (
+                    objects, dst_container, key))
+            self.logger.error('ETags: destination -- %s; source -- %s' % (
+                resp.headers['etag'], total_etag))
+            resp = self.delete_object(key, bucket=dst_container)
+            if not resp.success and resp.status != 404:
+                self.logger.error('Failed to remove corrupt segment: %s/%s' % (
+                    dst_container, key))
+                resp.reraise()
```
mdegerne: I thought key here referred to the combined object, not the individual segment. If so, this message should be reworded.

timuralp: You're right! Fixing.

timuralp: Reworded this to "Failed to remove corrupt combined object: <container/key>". The `key` is the combined object, which in our case usually is the segment, but you're right that that might be confusing.

s3_sync/sync_swift.py
```diff
@@ -693,25 +849,32 @@ def _upload_segment_worker(self, work_queue, internal_client,
                             (dst_container, resp.to_wsgi()))
                     errors.append(work)
```
thiagodasilva: I think this line could be removed or moved under the `if not resp.success` block, it causes a lot of logging if we need to create a container on the remote cluster

test/integration/test_s3_sync.py
```diff
@@ -1097,8 +1101,115 @@ def test_swift_slo_sync(self):
             clear_swift_container(conn, container)
             clear_swift_container(conn, container + '_segments')
 
+    def test_swift_slo_sync_resize_segments(self):
+        content = 'A' * 1024 + 'B' * 1024 + 'C' * 512
+        key = 'test_slo'
+        mapping = dict(self.swift_sync_mapping())
+        mapping['min_segment_size'] = 2000
+        mapping['convert_dlo'] = True
```
thiagodasilva: it's a bit weird that `convert_dlo` option is required to convert this slo. I guess we expect the feature to be used mainly for converting small DLO segments into SLOs, but this tests still shows this awkard requirement for an slo object

test/integration/test_s3_sync.py
```diff
@@ -1239,6 +1356,124 @@ def test_swift_dlo_sync(self):
...
+        slo_etag = hashlib.md5(''.join(
+            [hashlib.md5(content[i] + content[i + 1]).hexdigest()
+             for i in range(0, len(content), 2)])).hexdigest()
+        self.assertEqual('"%s"' % slo_etag, headers['etag'])
+
+        _, manifest_blob = self.remote_swift(
+            'get_object', mapping['aws_bucket'], key,
+            query_string='multipart-manifest=get')
+        manifest = json.loads(manifest_blob)
+        self.assertEqual(len(content) / 2, len(manifest))
```
thiagodasilva: I'm not sure I understand this statement, `content` is the body of the object, why half the `content` is the same size as the SLO `manifest`? what am I missing?

s3_sync/sync_swift.py
timuralp: Agreed. Done.

timuralp: On second thought, no, that was correct. The reason is that we need to externalize the error (the missing segments container) and there is no way to do that without adding the object to the errors list. However, maybe we should check the container existence before the upload? I think that might actually make sense. What do you think?

test/integration/test_s3_sync.py
timuralp: Agreed. I don't know how to do this better, though, because how would we handle the case of setting minimum segment size and _not_ converting DLO to SLO when we encounter a DLO? I guess we could name all the segments ourselves like we do today and set the prefix in the same fashion? I think we could open a bug/feature request for that.

test/integration/test_s3_sync.py
timuralp: `len(manifest)` is the number of segments in the manifest. `len(content)` is the number of original segments.

thiagodasilva: duh, my brain was mush last night

test/integration/test_s3_sync.py
thiagodasilva: Yep, agreed not an easy solution.  My fear of changing object names comes from the thinking that users could have built in clients that access segments directly?? what's the likelihood of that happening?

s3_sync/sync_swift.py
thiagodasilva: Yep, I like the idea of checking the container existence before the upload, I believe that's how the swift client (cli) behaves, seems sensible.

PR #260, opened 2019-02-08
Delete segments when archive

thiagodasilva:
Need to add tests


s3_sync/sync_container.py
```diff
@@ -191,6 +193,77 @@ def handle_container_metadata(self, metadata_dict, db_id):
...
+
+        if failed_deletes > 0:
+            self.logger.error("Failed to delete %s segments of DLO %s/%s" % (
+                failed_deletes, self._container, obj))
+
+    def _delete_slo_segments(self, swift_client, obj):
+        headers, manifest = _get_internal_manifest(
+            self._account, self._container, obj, swift_client, {})
+        self.logger.debug("JSON manifest: %s" % str(manifest))
+        failed_deletes = 0
```
timuralp: I'd keep a list of failed object names here as well.

s3_sync/sync_container.py
```diff
@@ -191,6 +193,77 @@ def handle_container_metadata(self, metadata_dict, db_id):
...
+        if failed_deletes > 0:
+            self.logger.error("Failed to delete %s segments of SLO %s/%s" % (
+                failed_deletes, self._container, obj))
+
+    def _delete_segments(self, swift_client, obj):
+        try:
+            metadata = swift_client.get_object_metadata(
+                self._account, self._container, obj)
+        except UnexpectedResponse as e:
+            if e.resp.status_int == 404:
```
timuralp: I realize this happens all over the place, but we could probably use `acceptable_statuses` and clean it up.

s3_sync/sync_container.py
```diff
@@ -191,6 +193,77 @@ def handle_container_metadata(self, metadata_dict, db_id):
...
+            except:
+                failed_deletes += 1
+
+        if failed_deletes > 0:
+            self.logger.error("Failed to delete %s segments of SLO %s/%s" % (
+                failed_deletes, self._container, obj))
+
+    def _delete_segments(self, swift_client, obj):
+        try:
+            metadata = swift_client.get_object_metadata(
```
timuralp: I wonder if we should maybe sink this into the providers to avoid making an extra HEAD request (and GET manifest and DLO listing in case of those objects). That would also make it easier for us to do the deletions in parallel (probably a must to make this viable for Pac-12 DLOs).

s3_sync/sync_container.py
```diff
@@ -215,16 +288,20 @@ def handle(self, row, swift_client):
                 # mitigate races where the object may be overwritten. We
                 # increment the offset to ensure that we never remove new
                 # customer data.
-                self.logger.debug("Creating a new TS: %f %f" % (
-                    meta_ts.offset, meta_ts.timestamp))
+
+                # handle large objects: delete segments first
+                if not self.retain_local_segments:
+                    self._delete_segments(swift_client, row['name'])
```
timuralp: If we fail to delete segments, it probably doesn't make sense to delete the manifest?

s3_sync/sync_swift.py
```diff
@@ -482,7 +483,7 @@ def update_metadata(self, swift_key, metadata, remote_metadata={},
             user_headers[dlo_etag_field] = remote_metadata.get(dlo_etag_field)
         self.post_object(swift_key, user_headers, container)
 
-    def _upload_objects(self, objects_list, internal_client):
+    def _upload_large_objects(self, objects_list, internal_client):
```
timuralp: I'd prefer not to make this change -- the internal method accepts a list of objects rather than large objects and doesn't really interrogate what they are.

s3_sync/utils.py
```diff
@@ -1131,6 +1131,23 @@ def get_dlo_prefix(swift_meta):
     return swift_meta[MANIFEST_HEADER]
 
 
+def _get_internal_manifest(acc, cont, obj, internal_client, swift_headers={}):
```
timuralp: Please drop the leading `_`, since it's not an internal function anymore.

s3_sync/sync_container.py
```diff
@@ -191,6 +193,77 @@ def handle_container_metadata(self, metadata_dict, db_id):
...
+        def _keys_generator():
+            internal_iterator = iter_internal_listing(
+                swift_client, self._account, container=s_container,
+                prefix=s_prefix)
+            for entry in internal_iterator:
+                if not entry:
+                    return
+                yield entry['name']
+
+        failed_deletes = 0
```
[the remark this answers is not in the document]
thiagodasilva: I considered a csv list, but then thinking about how we could dump a big list (1000 segments) wouldn't provide to be that helpful, wdyt?

s3_sync/sync_container.py
```diff
@@ -191,6 +193,77 @@ def handle_container_metadata(self, metadata_dict, db_id):
             self.logger.debug(
                 ''.join(traceback.format_tb(post_resp.exc_info[2])))
 
+    def _delete_object(self, swift_client, acc, cont, obj, hdrs={}):
+        try:
+            swift_client.delete_object(
+                acc, cont, obj, headers=hdrs)
+        except UnexpectedResponse as e:
+            if e.resp.status_int in (404, 409):
```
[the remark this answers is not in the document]
thiagodasilva: done

s3_sync/sync_container.py
```diff
@@ -191,6 +193,77 @@ def handle_container_metadata(self, metadata_dict, db_id):
...
+            internal_iterator = iter_internal_listing(
+                swift_client, self._account, container=s_container,
+                prefix=s_prefix)
+            for entry in internal_iterator:
+                if not entry:
+                    return
+                yield entry['name']
+
+        failed_deletes = 0
+        for obj in _keys_generator():
```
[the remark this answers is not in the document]
thiagodasilva: done

s3_sync/utils.py
thiagodasilva: done

s3_sync/sync_swift.py
thiagodasilva: doesn't the call to `upload_segment_worker` suggest that this method would only be used for large objects with segments? can't see it being used for anything else, thoughts?

s3_sync/sync_container.py
thiagodasilva: this is another one I struggled with just because we could ended up in a situation where some segments might have been deleted, but not all. I was trying to weigh which one would be worse, not sure...

s3_sync/sync_container.py
thiagodasilva: good idea

s3_sync/sync_container.py
thiagodasilva: I agree with the need to make the deletions in parallel, but isn't the internal client still only one instance? Are you suggesting we create a internal_client pool (like in the crawler) or that we just use a pool of workers while using the same internal client? what if the crawler passed in the ic_pool as opposed to passing in the ic instance itself?

s3_sync/sync_swift.py
timuralp: I'd say we should rename the `upload_segment_worker` method, though. None of these methods rely on anything about large objects -- they don't inspect manifests or do additional handling. This is a pretty minor point that I would leave as is, to be honest.

s3_sync/sync_container.py
timuralp: If we delete the manifest, we'll never find those segments again. Whereas if we preserve the manifest and emit an error, we will re-attempt to delete them next time.

s3_sync/sync_container.py
timuralp: Internal client doesn't keep any internal state, as far as I can tell, and we rely on making multiple requests in parallel with 1 instance (that's how SLO/DLO uploads work today already).

s3_sync/sync_container.py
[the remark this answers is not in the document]
timuralp: We could dump it at debug level, at least. You can report the list length at info level. Otherwise, we have no idea what deletes actually failed and no way to obtain that in the future.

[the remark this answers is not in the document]
thiagodasilva: sounds good, setting at debug level now, but we can change later..

s3_sync/sync_container.py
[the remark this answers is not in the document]
thiagodasilva: done

s3_sync/sync_container.py
thiagodasilva: done

s3_sync/sync_container.py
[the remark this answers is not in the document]
thiagodasilva: done

s3_sync/sync_container.py
thiagodasilva: sounds good, will make the change.

s3_sync/sync_container.py
thiagodasilva: done

s3_sync/sync_swift.py
thiagodasilva: keeping as _upload_objects

s3_sync/base_sync.py
```diff
@@ -339,3 +342,122 @@ def _full_name(self, key):
         return u'%s/%s/%s' % (self.account, self.container,
                               key if isinstance(key, unicode)
                               else key.decode('utf-8'))
+
+    def _delete_objects(self, objects_list, swift_client):
+        '''
+        Initialize a task queue to delete objects in parallel
+        '''
+        work_queue = eventlet.queue.Queue(self.SLO_QUEUE_SIZE)
```
timuralp: Do we need to pull in these constants into the base_sync class? This probably fails with S3?

test/integration/test_s3_sync.py
```diff
@@ -809,6 +809,110 @@ def test_swift_archive_slo_w_remote_delete_at(self):
         # NOTE: we only record the SLO as being uploaded, not the segments
         self._assert_stats(mapping, 1, 'copied_objects')
 
+    def test_swift_archive_slo_w_delete_segments(self):
```
timuralp: We should add a similar SLO test when targeting S3.

s3_sync/base_sync.py
thiagodasilva: they are defined in BaseSync, lines 112,113

timuralp: Oh, sorry, I missed that.

test/integration/test_s3_sync.py
thiagodasilva: done

PR #266, opened 2019-02-14
Change RuntimeError to RetryError

thiagodasilva:
Just trying to avoid the double logging when there's an error to
to upload segments. The worker thread already logs once, and with
RuntimeError, container crawler was logging again. Since all we really
want is for a retry later, changed error to RetryError.


timuralp:
Maybe we can improve how we handle these workers' error messages in the future. This seems OK as a stop-gap measure.

PR #272, opened 2019-02-26
Add statsd output on the migrator

ctennis:


s3_sync/stats.py
```diff
@@ -26,13 +26,20 @@ def update(self, **kwargs):
...
-        self.bytes_copied += bytes_copied
+        if copied > 0:
+            self.copied += copied
+            self.statsd_increment('copied_objects', copied)
+        if scanned > 0:
+            self.scanned += scanned
+            self.statsd_increment('scanned', scanned)
+        if bytes_copied > 0:
+            self.bytes_copied += bytes_copied
+            self.statsd_increment('bytes', bytes_copied)
```
timuralp: We should track this metric like we do with sync/lifecycle: measuring bandwidth, rather than reporting bytes transferred after the fact. The migrator currently can't do that, as we don't get that information from the InternalClient, however we could wrap the body to do so.

s3_sync/stats.py
```diff
@@ -26,13 +26,20 @@ def update(self, **kwargs):
...
         self.scanned = 0
         self.bytes_copied = 0
 
     def _update_stats(self, copied=0, scanned=0, bytes_copied=0):
-        self.copied += copied
-        self.scanned += scanned
-        self.bytes_copied += bytes_copied
+        if copied > 0:
+            self.copied += copied
+            self.statsd_increment('copied_objects', copied)
```
timuralp: Making statsd calls here doesn't seem right. I would move it back into the migrator. Also, it's holding a lock while making this system call.

s3_sync/shunt.py
```diff
@@ -283,15 +283,16 @@ def __call__(self, env, start_response):
     def iter_remote_objects(
             self, sync_profile, per_account, marker, limit, prefix, delimiter):
         provider = create_provider(sync_profile, max_conns=1,
-                                   per_account=per_account)
+                                   per_account=per_account, logger=self.logger)
```
timuralp: Is this change related to the stats? Can we do it separately?

s3_sync/migrator.py
```diff
@@ -1252,15 +1287,18 @@ def process_migrations(migrations, migration_status, internal_pool, logger,
...
         cycle_start = time.time()
         process_migrations(migrations, migration_status, internal_pool, logger,
                            items_chunk, workers, selector,
-                           segment_size)
+                           segment_size, statsd_client)
         elapsed = time.time() - cycle_start
         naptime = max(0, poll_interval - elapsed)
         msg = 'Finished cycle in %0.2fs' % elapsed
+        if statsd_client:
+            statsd_client.timing('1space.migrator.cycle_time', elapsed * 1000)
```
timuralp: Looks like forgot to update the metric name (probably just `cycle_time`).

s3_sync/migrator.py
```diff
@@ -1314,8 +1352,16 @@ def main():
     migrations = conf.get('migrations', [])
     migration_status = Status(migrator_conf['status_file'])
 
+    if conf.get('statsd_host'):
```
timuralp: Let's use `migrator_conf` here, as it allows to have a separate statsd prefix for sync/lifecycle vs migrations.

s3_sync/migrator.py
```diff
@@ -336,7 +338,34 @@ def __init__(self, config, status, work_chunk, workers, swift_pool, logger,
         self.provider = None
         self.gthread_local = eventlet.corolocal.local()
         self.segment_size = segment_size
+        self.statsd_client = statsd_client
         self.handled_containers = []
+        self.start_time = time.time()
+
+        statsd_prefix_parts = [
```
timuralp: This looks pretty similar to what we're doing in `SyncContainer`. I wonder if we should subclass `pystatsd.statsd.Client`, add this to the constructor, and implement the `increment` and `timing` methods to use it. Or use composition, which would allow the `if` conditions to be sunk in there too. What do you think?

ctennis: reworked

s3_sync/migrator.py
```diff
@@ -505,6 +532,7 @@ def _next_pass(self):
     def check_errors(self):
         while not self.errors.empty():
             container, key, err = self.errors.get()
+            self.stats_reporter.increment('error_count', 1)
```
timuralp: We could probably call it once with the length of `self.errors()`?

s3_sync/migrator.py
```diff
@@ -1183,6 +1213,9 @@ def _upload_object(self, work, preserve_timestamp=True):
         if result.status_int == 201:
             self.gthread_local.uploaded_objects += 1
             self.gthread_local.bytes_copied += size
+            self.stats_reporter.increment('copied_objects', 1)
+            if size > 0:
+                self.stats_reporter.increment('bytes', size)
```
timuralp: I worry about disambiguating in-flight bytes that we'll definitely add later vs the more coarse grained value here. Although, we can probably replace one with the other later.

s3_sync/sync_container.py
```diff
@@ -243,16 +238,13 @@ def __str__(self):
...
 
     def instance(self, settings, per_account=False):
-        if 'statsd_host' in self.config:
-            statsd_client = pystatsd.statsd.Client(
-                self.config['statsd_host'],
-                self.config.get('statsd_port', 8125),
-                self.config.get('statsd_prefix'))
-        else:
-            statsd_client = None
+        stats_factory = StatsReporterFactory(
```
timuralp: I wouldn't drop the `statsd_host` part being optional, but maybe we can make a NOOP reporter to avoid the if/else clauses all over the place in the users of the factory.

s3_sync/migrator.py
```diff
@@ -337,6 +339,27 @@ def __init__(self, config, status, work_chunk, workers, swift_pool, logger,
         self.gthread_local = eventlet.corolocal.local()
         self.segment_size = segment_size
         self.handled_containers = []
+        self.start_time = time.time()
+        self.stats_factory = stats_factory
```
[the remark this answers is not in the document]
ctennis: Unfortunately I think we do.  That's because later (around like 426) in the _reconcile_containers method we may have to generate instances from the factory as this method loops over containers when we're in the "*/" configuration.  Since the container names aren't known until this point, and because our stats names include the container in them.

s3_sync/migrator.py
```diff
@@ -337,6 +339,27 @@ def __init__(self, config, status, work_chunk, workers, swift_pool, logger,
         self.gthread_local = eventlet.corolocal.local()
         self.segment_size = segment_size
         self.handled_containers = []
+        self.start_time = time.time()
```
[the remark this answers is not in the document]
ctennis: Going to remove this line.  It was part of some timing work I was doing, but it will go in another PR later.

s3_sync/migrator.py
ctennis: Going to remove this line.  Will work on error_count stuff in a future PR.

s3_sync/migrator.py
```diff
@@ -337,6 +339,27 @@ def __init__(self, config, status, work_chunk, workers, swift_pool, logger,
...
+        self.stats_factory = stats_factory
+
+        statsd_prefix = self._build_statsd_prefix(self.config)
+        self.stats_reporter = self.stats_factory.instance(statsd_prefix)
+
+    def _build_statsd_prefix(self, config):
```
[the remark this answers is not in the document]
ctennis: Nope, good catch.

s3_sync/sync_container.py
ctennis: I think that's what this does?  The setting is still optional, but now we create the factory and the factory is aware that the host isn't set, so calls to the factory/instance no-op inside of the instance itself vs. doing the if/elses everywhere as you noted.

s3_sync/migrator.py
ctennis: Not sure I understand. In this method, it looks to me like when we get to this point, the object has been fully transferred?

s3_sync/migrator.py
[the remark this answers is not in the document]
timuralp: I see -- makes sense.

s3_sync/migrator.py
timuralp: Don't worry about it! I was talking about getting in-flight bandwidth stats, but we don't have to do it in this review. I might be able to help with that subsequently.

s3_sync/sync_container.py
timuralp: I'm sorry -- I missed that clause! Never mind.

PR #274, opened 2019-03-05
Allow for empty "Body" or "Buckets" in SyncS3.

timuralp:
Boto may return an empty "Body" or "Buckets" keys when there is no
actual content. In case of a 0-sized object, this doesn't affect us, as
we will return an empty string. However, when returning results from an
account with no buckets, we should still return an empty list (as
opposed to an empty string).


timuralp:
@mdegerne right, if there are no buckets, it returns an empty list for the `Buckets` key, which evaluates to `False` and we end up with `iter([''])` in the original code.

PR #279, opened 2019-03-09
Update migrator transfer stats after read().

timuralp:
The goal of the patch is to provider an in-progress view of the bandwidth when copying objects in the migrator.

### DIFF
--- a/s3_sync/migrator.py
+++ b/s3_sync/migrator.py
@@ -17,7 +17,6 @@ import eventlet.corolocal
 import eventlet.pools
 eventlet.patcher.monkey_patch(all=True)
 
-from collections import namedtuple
 import datetime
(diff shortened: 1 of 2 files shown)


timuralp:
@mdegerne @ctennis actually added tests for this in his prior PR (so they didn't show up in this change). I was only changing the bytes tracking to reflect ongoing bandwidth usage, rather than reporting it at the end of the transfer.

timuralp:
I'm going to push this in, but as two separate commits without squashing them.

timuralp:
Merged via ee1194e17f9a5dd9ec15401f71df6640ce7236bb and 085870e5d5fbf0d5425394be630cd79d0423d1ae.

PR #280, opened 2019-03-11
Fix a double-counting issue for manifests.

timuralp:
We have an issue where SLO manifests are double-counted for the migrator
stats purposes. Added tests to validate the migrator stats are kept
correctly for regular objects, DLOs, and SLOs.


test/integration/test_migrator.py
```diff
@@ -357,6 +362,13 @@ def _check_segments(prefix):
              'bytes': (segment_len * parts_count * 2) + 2030
              })
 
+        migration_status = status.get_migration(migration)
+        self.assertEqual(22, migration_status['moved_count'])
+        # SLO segments are not *scanned*
+        self.assertEqual(parts_count + 2, migration_status['scanned_count'])
+        self.assertEqual(parts_count * segment_len * 2 + 2030,
```
mdegerne: I know from context that 2030 is the manifest size, but I wonder if it will be obvious looking at this code a year from now.

timuralp:
@mdegerne you're right. Fixed.

PR #282, opened 2019-03-14
Storage policies

mdegerne:
Storage policy inclusion for both sync and migrate. I would like to keep these as separate commits, if possible.


s3_sync/sync_swift.py
```diff
@@ -866,8 +870,11 @@ def _upload_segment_worker(self, work_queue, internal_client,
                     # Creating a container may take some (small) amount of time
                     # and we should attempt to re-upload in the following
                     # iteration
+                    headers = {}
+                    if self.storage_policy:
+                        headers['X-Storage-Policy'] = self.storage_policy
                     resp = self._call_swiftclient(
-                        'put_container', dst_container, None)
+                        'put_container', dst_container, None, headers=headers)
```
timuralp: We don't have to do it as part of this PR, but it'd be great to factor out the container creation into a separate function.

test/integration/test_s3_sync.py
```diff
@@ -161,6 +161,102 @@ def test_swift_sync_account_slo(self):
...
+        conn.put_container(container)
+        conn.put_object('segments', 'part1', content[:1024])
+        conn.put_object('segments', 'part2', content[1024:])
+        conn.put_object(container, key,
+                        json.dumps(manifest),
+                        query_string='multipart-manifest=put')
+
+        crawler.run_once()
+
+        hdrs = target_conn.head_object(
```
timuralp: I would remove all of these assertions (since we already have a test that copies the objects around), but add a check for the storage policy. I don't see us checking that the right storage policy was used anywhere. We should check the container with the manifest and the segments containers.

mdegerne:
TBH, I had punted on twiddling the container because I wasn't sure how much work that was. If we get a second storage policy it makes a bunch of this easier to verify. :)

timuralp:
@mdegerne alright. Would you mind looking into what it would take to add a second storage policy? I suspect this shouldn't be too hard. It can be replicated and exactly like the default one, but with a different name (and index)

timuralp:
Merged via 193f972f3ed6a0f8d3cdd88764e62203a444faa6 and f979c518b9ace5dbc6604b5e8e61f4c2bc572bb0 (squashed the last 4 commits involving verification, sync/lifecycle changes, and tests together).

PR #283, opened 2019-03-18
New Crawler interface to obtain container stats.

timuralp:
Adds support for the new container_crawler.Crawler interface that allows
for the clients to examine the container stats (as well as container
metadata).


timuralp:
That's the plan! I'll update this PR once new container crawler is available with the corresponding patch merged there.

PR #285, opened 2019-04-01
Ignore File Connector etag when converting headers

thiagodasilva:
We had a conditional to ignore ProxyFS etag headers, just
extending to also ignore File Connector etags.


s3_sync/utils.py
```diff
@@ -1137,7 +1138,8 @@ def convert_to_local_headers(headers, remove_timestamp=True):
     # restoring from Swift.
     if 'x-timestamp' in put_headers and remove_timestamp:
         del put_headers['x-timestamp']
-    if put_headers['etag'].startswith('"%s' % PFS_ETAG_PREFIX):
+    if put_headers['etag'].startswith('"%s' % PFS_ETAG_PREFIX) or \
+       put_headers['etag'].endswith(FC_ETAG):
```
timuralp: Would we not use the same format as the pfs etag (i.e. use the startswith comparison vs endswith above)? What's the reason for appending it vs including a prefix?

timuralp: Also, it should probably return a quoted ETag to be RFC compliant (unlike Swift for regular objects :p)

thiagodasilva: re: format of etag. I can make the change to look more like pfs etag. At one point I thought of having a etag like '000-fcetag' just in case in the future we wanted to put any other information before the ``- `` but I'm happy to just change it ot ``fcetag``.

re: quoted ETag, In the File Connector the etag is set at the DiskFile layer, so even when I quote it, the Swift proxy removes it. The File Connector runs the upstream proxy basically unchanged, so I'd prefer to not change that right now...

PR #286, opened 2019-04-03
Get valid storage policy name direct from swift

mdegerne:
Improve the test by grabbing a real storage policy name (so that it can
run on other environments).


thiagodasilva:
To add to what @timuralp said above, some of this code here could be helpful: https://github.com/openstack/swift/blob/master/test/unit/__init__.py#L91

timuralp:
We can improve the unit tests later (to not assume `/etc/swift/swift.conf` existence).

PR #287, opened 2019-04-08
Fix the syslog output.

timuralp:
Changes syslog messages to be LF separated, not including a NULL
character, and include the hostname. This required amending the existing
SysLogHandler in python (which has additional options in python3, so we
could avoid patching it in the future).


s3_sync/daemon_utils.py
```diff
@@ -22,12 +21,72 @@
...
+       message priority (a combination of facility and log level). We resolve
+       this by appending "\n" to the message, as opposed to NULL when using TCP
+       (SOCK_STREAM) sockets.
+       NOTE: In python3, the appended NULL character is made optional, but a
+       custom formatter is still required to append "\n" to every message, as
+       otherwise they are still stuck together.
+       NOTE: The logging.Handler class is not thread safe if used with eventlet
+       across different pthreads (however, we currently run in one pthread).
+    '''
+    def emit(self, record):
```
thiagodasilva: compared with code here: https://github.com/python/cpython/blob/master/Lib/logging/handlers.py#L910 and it looks good.

PR #288, opened 2019-04-09
Handle casing in header names.

timuralp:
Apparently, we can get back title-cased headers from boto3, which breaks
our object metadata comparison. The patch uses the requests
CaseInsensitiveDict to compare the header values for the object metadata
from S3.


timuralp:
@mdegerne there is a TODO left over from the last iteration of this code regarding the `remote_delete_after` setting being removed. Do you remember the context of it? What do we need to do? It would be great to clean that up as well.

mdegerne:
I don't recall the context for that TODO. :( I'm sure it seemed obvious at the time.

PR #295, opened 2019-05-16
Migrate big file connector files as SLO

thiagodasilva:
Fixed support for migrating large objects that might not have
an etag (i.e., File Connector) on first read.


s3_sync/migrator.py
```diff
@@ -1105,16 +1105,20 @@ def _upload_manifest_from_segments(
...
         headers['X-Static-Large-Object'] = 'True'
         work = UploadObjectWork(container, key, manifest, headers, aws_bucket)
         self._upload_object(work)
 
     def _migrate_as_slo(self, aws_bucket, container, key, resp, put_headers):
         data = resp.body
         content_length = int(resp.headers['Content-Length'])
-        remote_etag = put_headers['etag']
-        del(put_headers['etag'])
+        if put_headers.get('etag', None):
```
timuralp:
```suggestion
        if put_headers.get('etag'):
```

PR #297, opened 2019-05-24
Fix syncing SLOs with segments from different containers

thiagodasilva:
Previously the upload slo code expected all SLO segments
to be stored in the same container (This is indeed the most common
use case) but the Swift API does allow for manifest to be made up
of segments stored in different containers.

This patch changes the SLO syncing code to handle manifests
containining segments stored in different containers.


s3_sync/sync_swift.py
```diff
@@ -836,23 +839,28 @@ def _upload_segment_worker(self, work_queue, internal_client,
...
                 kwargs['segment'] = True
                 upload_func = self._upload_object
                 error_msg = 'Failed to upload segment %s/%s/%s' % (
                     self.account, container, obj)
             else:
-                container, obj_list, obj = work
-                dst_container = self.remote_segments_container(container)
-                args = [container, dst_container, obj, obj_list,
-                        internal_client]
+                # When combining slo segments, it is possible that segments
```
timuralp: I don't think we need this comment, since we always put SLO segments into the one segments container for the main container (hence `remote_segments_container` being used everywhere).

test/integration/test_s3_sync.py
```diff
@@ -1295,16 +1295,20 @@ def test_swift_archive_slo_restore(self):
...
     def test_swift_slo_sync(self):
-        content = 'A' * 2048
+        content = 'A' * 3072
         key = 'test_slo'
         mapping = self.swift_sync_mapping()
         crawler = utils.get_container_crawler(mapping)
         manifest = [
             {'size_bytes': 1024, 'path': '/segments/part1'},
-            {'size_bytes': 1024, 'path': '/segments/part2'}]
+            {'size_bytes': 1024, 'path': '/seg2/part2'},
```
timuralp: I'd prefer a new test for the segments in different containers, rather than repurposing this test.

thiagodasilva: done

PR #298, opened 2019-05-25
Allow setting the Storage Class for AWS S3 objects.

timuralp:
AWS S3 supports numerous storage classes for data at various price
points. The objects can be PUT into the storage class using the
StorageClass parameter in boto3 (including GLACIER and DEEP_ARCHIVE).

The patch allows a new configuration option when specifying 1space
mappings: storage_class. Valid values are those allowed by boto3/Amazon.
This optional setting is only available for AWS S3 endpoints.


thiagodasilva:
@timuralp I just quickly read the patch and it looks good. I haven't run any manual testing and I'm not sure about the failed Travis test, it looks unrelated but I don't remember us having tests that fail intermittently.

PR #302, opened 2019-06-06
Sync large objects to File Connector.

thiagodasilva:
Re-using _upload_combined_objects method to upload xLO objects
as as single object to a File Connector.

Added an additional remote call to /info to determine if remote
cluster supports SLO and use this information to dynamically
determine if it's a File Connector endpoint. In the future we could
added an option to configuration file that specifies type of cluster.


onespace/sync_swift.py
```diff
@@ -488,36 +496,60 @@ def _upload_object(self, src_container, dst_container, key,
...
-                    self.update_metadata(key, metadata,
-                                         remote_metadata=remote_meta,
-                                         bucket=dst_container)
-                    # TODO: Update segments' X-Delete-At headers if
-                    # remote_delete_after is applied/updated/removed.
-                    return self.UploadStatus.POST
-                return self.UploadStatus.NOOP
-            return self._upload_dlo(key, internal_client, metadata, req_hdrs,
-                                    stats_cb=stats_cb)
+            # TODO: this is a bit of hack to determine if remote cluster
```
timuralp: I think it wouldn't be a huge leap to move to having a separate provider for the file connector. That would simplify some of the branching here and the discovery we have to do. What do you think? Would you be up for exploring what that might look like?

PR #303, opened 2019-06-07
WIP: Add File Connector integration tests

thiagodasilva:


test/integration/test_migrator.py
```diff
@@ -2053,3 +2053,57 @@ def test_s3_mpu_migration(self):
         clear_swift_container(conn, container + '_segments')
         conn.delete_container(container)
         conn.delete_container(container + '_segments')
+
+    def test_file_connector_migration(self):
+        # Run migrator with custom migration definition
+        fc_mig = self.swift_migration()
```
mdegerne: I find it confusing to do this, since you are only using the account and container. It would be clearer if you explicitly extracted those here and then never used fc_mig again. Alternatively, you could just specify the account and/or container explicitly here.

scripts/ensure_file_connector_started
```diff
@@ -0,0 +1,33 @@
...
+
+set -e
+
+# prepare fcdata dir
+rm -rf /tmp/fcdata
+mkdir -p /tmp/fcdata/migration
+mkdir -p /tmp/fcdata/sync
+
+for i in {0..2}; do echo hello obj$i > /tmp/fcdata/migration/obj$i; done
+bash -c "printf 'A%.0s' {1..$((2*1024))} > /tmp/fcdata/migration/slo_obj"
```
mdegerne: I am not convinced that I like the idea of creating the object that is used for integration testing so far removed from the integration test itself. It makes reading the test(s) that much more difficult when you have to go and search for where the object is created.

thiagodasilva: Agree. I was trying to test functionality in the file connector itself where it creates the metadata directory the first time the object is read/written over object API, but that's something we can test in the File Connector repo itself. For now I will just create the test data using the object API for each test individually.

test/integration/test_migrator.py
thiagodasilva: sounds good, let me try to clean it up a bit.

